PR #3, opened 2025-12-06
Second wave of compilation fixes

kipcode66:


src/d/d_camera.cpp
```diff
@@ -4223,13 +4223,13 @@ bool dCamera_c::lockonCamera(s32 param_0) {
...
-                fopAcM_getNameString(mpPlayerActor, NULL);
-                dDbVw_Report(0x1e0, 0x109, "%s", NULL);
-            }
+#ifndef NDEBUG
+        if (mCamSetup.CheckFlag(0x8000)) {
+            //char name[28];
+            fopAcM_getNameString(mpPlayerActor, NULL);
+            dDbVw_Report(0x1e0, 0x109, "%s", NULL);
         }
+#endif
```
TakaRikka: for changes like these, did you test building decomp to make sure it doesnt break matching?

src/d/d_com_static.cpp
```diff
@@ -272,8 +272,8 @@ const char* daSetBgObj_c::getArcName(fopAc_ac_c* i_this) {
 s16 const daObjMovebox::Act_c::M_dir_base[4] = {
     0x0000,
     0x4000,
-    0x8000,
-    0xC000,
+    static_cast<s16>(0x8000),
+    static_cast<s16>(0xC000),
```
TakaRikka: with these, if you can test to see if it matches using a signed value instead, maybe we dont need to specifically cast it

src/d/d_ev_camera.cpp
```diff
@@ -995,7 +995,7 @@ bool dCamera_c::transEvCamera(int param_1) {
             trans->mEye = (*ptr1).mEye;
             trans->mFovy = (*ptr1).mFovy;
             trans->field_0x3c = (*ptr1).mBank;
-            if ((*ptr1).mBank != 0.0f) {
+            if ((*ptr1).mBank.Degree() != 0.0f) {
```
TakaRikka: according to debug these inlines werent used originally afaict.

src/d/d_file_select.cpp
```diff
@@ -485,7 +485,7 @@ void dFile_select_c::dataSelectInAnmSet() {
             }
         }
 
-        (char*)pSave += 0xa94;
+        pSave = (dSv_save_c*)((u8*)pSave + 0xa94);
```
TakaRikka, line 488: same here about testing decomp build

src/d/d_file_select.cpp
```diff
@@ -4096,5 +4096,7 @@ void dFile_select3D_c::toItem3Dpos(f32 param_0, f32 param_1, f32 param_2, cXyz*
 #pragma pop
 
 void dFile_select3D_c::calcViewMtx(Mtx param_0) {
-    cMtx_lookAt(param_0, &cXyz(0.0f, 0.0f, -1000.0f), &cXyz::Zero, &cXyz(0.0f, 1.0f, 0.0f), 0);
+    auto pos1 = cXyz(0.0f, 0.0f, -1000.0f);
+    auto pos2 = cXyz(0.0f, 1.0f, 0.0f);
```
TakaRikka: these also need to fixed for decomp so you can try seeing if it matches, though we don't use auto in decomp

src/d/d_menu_collect.cpp
```diff
@@ -2579,7 +2579,9 @@ void dMenu_Collect3D_c::toItem3Dpos(f32 param_0, f32 param_1, f32 param_2, cXyz*
 #pragma pop
 
 void dMenu_Collect3D_c::calcViewMtx(Mtx param_0) {
-    mDoMtx_lookAt(param_0, &cXyz(0.0f, 0.0f, -1000.0f), &cXyz::Zero, &cXyz(0.0f, 1.0f, 0.0f), 0);
+    auto pos1 = cXyz(0.0f, 0.0f, -1000.0f);
+    auto pos2 = cXyz(0.0f, 1.0f, 0.0f);
```
TakaRikka: same thing here

src/d/d_meter2.cpp
```diff
@@ -3088,9 +3088,9 @@ static leafdraw_method_class l_dMeter2_Method = {
 };
 
 extern msg_process_profile_definition g_profile_METER2 = {
-    -3,
+    static_cast<unsigned int>(-3),
```
TakaRikka: we have some enums for these that'd probably be better, should be able to copy them from another g_profile example

TakaRikka (commented):
overall it looks fine, but i'd like to double check that most of the code changes still match when building decomp. once we start implementing port specific stuff i think it's fine to deviate away, but in places where we can keep matching while fixing compilation issues i think it'd be good to do it the proper way

PR #5, opened 2025-12-22
Add missing include directories

kipcode66:


CMakeLists.txt
```diff
@@ -1343,9 +1345,10 @@ source_group("dolzel" FILES ${DOLZEL_FILES} ${Z2AUDIOLIB_FILES} ${SSYSTEM_FILES}
 source_group("dusk" FILES ${DUSK_FILES})
 
 add_library(game SHARED ${DOLZEL_FILES} ${Z2AUDIOLIB_FILES} ${SSYSTEM_FILES} ${JSYSTEM_FILES} ${REL_FILES} ${DUSK_FILES})
-target_compile_definitions(game PRIVATE TARGET_PC VERSION=0)
+target_compile_definitions(game PRIVATE TARGET_PC VERSION=0 NDEBUG=1 NDEBUG_DEFINED=1 DEBUG_DEFINED=0)
 # TODO: version handling for res includes
-target_include_directories(game PRIVATE include assets/GZ2E01 src/PowerPC_EABI_Support/MSL/MSL_C++/MSL_Common/Include src/PowerPC_EABI_Support/MSL/MSL_C/MSL_Common/Include src/PowerPC_EABI_Support/MSL/MSL_C/MSL_Common_Embedded/Math/Include)
+set(DUSK_TP_VERSION GZ2E01)
+target_include_directories(game PRIVATE include src assets/${DUSK_TP_VERSION} ${CMAKE_BINARY_DIR}/../${DUSK_TP_VERSION}/include src/PowerPC_EABI_Support/MSL/MSL_C++/MSL_Common/Include src/PowerPC_EABI_Support/MSL/MSL_C/MSL_Common/Include src/PowerPC_EABI_Support/MSL/MSL_C/MSL_Common_Embedded/Math/Include)
```
TakaRikka, line 1351: wonder if the powerpc eabi stuff should be removed in favor of linking the normal stdlib? if thats gonna need a lot more cleanup then can save it for a future pr, but something to think about

kipcode66: I was thinking that we get it to compile at all first. Then we can look at changing the standard lib stuff once we have something that compiles.

PR #7, opened 2026-02-07
Fix linker errors.

crowell:
Stub out all unused functions, separated by #pragma marks for easier viewing.

this still has linker errors due to missing symbols


src/dusk/m_Do_ext_dusk.cpp
```diff
@@ -0,0 +1,565 @@
+// XXX Ripped these from a DEBUG block, rather than changing defines.
```
TakaRikka: i think its probably better to just remove the debug define for these in the main file

crowell: done 👍

src/d/actor/d_a_e_dd.cpp
```diff
@@ -336,7 +336,7 @@ static BOOL way_bg_check(e_dd_class* i_this, f32 param_2) {
     return FALSE;
 }
 
-u8 hio_set;
+static u8 hio_set;
```
TakaRikka, line 339: all of these changes where static is being added would probably be good to add to decomp, and then we'll pull it in here (assuming it doesnt break matching, which i wouldnt imagine it would)

PJB3005:
Can this be closed now that #9 is merged?

PR #8, opened 2026-02-08
update build instructions to show how to use configure.py

crowell:


TakaRikka (approved):

README.md
```diff
@@ -1,4 +1,18 @@
-build
+build (Win32)
+
 ```bash
+cp /path/to/twilight-princess.ciso orig/GZ2E01/.
```
TakaRikka, line 4: not a major thing but its probably better to leave it as `.iso` rather than `.ciso`

PR #14, opened 2026-02-24
Compile majority of JSystem (& SSystem) with debug asserts enabled

PJB3005:
I went through and enabled JSystem parts that seemed to not cause linker errors when compiled in DEBUG


PJB3005:
I love CMake (this breaks the build?)

PR #15, opened 2026-02-25
fix build on clang/mac

crowell:


src/d/d_meter2_info.cpp
```diff
@@ -347,6 +347,7 @@ void dMeter2Info_c::decMsgKeyWaitTimer() {
     }
 }
 
+#if !TARGET_PC
```
TakaRikka: is there a reason these are being left out here? these are quite important for the game

TakaRikka (commented):
conflict also needs to be handled

src/d/d_meter2_info.cpp
crowell:
```jeff@wasteland dusk ( wip/mac$!)$ git grep "void dMeter2Info_c::getString("
src/d/d_meter2_info.cpp:void dMeter2Info_c::getString(u32 i_stringID, char* o_string, JMSMesgEntry_c* i_msgEntry) {
src/dusk/stubs.cpp:void dMeter2Info_c::getString(u32 i_stringID, char* o_string, JMSMesgEntry_c* i_msgEntry) {```

there's a copy in 2 places, I'll remove the one in stubs instead? There were a few places with multiple copies of the same symbol.

crowell:
rebased/fixed conflicts

PR #23, opened 2026-02-25
Fix broken J3DTexture assert

PJB3005:
Likely bad decomp according to Jasper

### DIFF
--- a/include/JSystem/J3DGraphBase/J3DTexture.h
+++ b/include/JSystem/J3DGraphBase/J3DTexture.h
@@ -19,7 +19,7 @@ private:
 
 public:
     J3DTexture(u16 num, ResTIMG* res) : mNum(num), unk_0x2(0), mpRes(res) {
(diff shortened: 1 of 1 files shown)


include/JSystem/J3DGraphBase/J3DTexture.h
```diff
@@ -19,7 +19,8 @@ class J3DTexture {
 
 public:
     J3DTexture(u16 num, ResTIMG* res) : mNum(num), unk_0x2(0), mpRes(res) {
-        J3D_ASSERT_NULLPTR(52, res && num);
+        // Changed assert from decomp: seems decomp is wrong?
+        J3D_ASSERT_NULLPTR(52, !!num == !!res);
```
TakaRikka: just use this, i already confirmed in the discord that this matches for shieldD
```cpp
J3D_ASSERT_NULLPTR(52, res != NULL || num == 0);
```

PR #27, opened 2026-02-27
fix more compile errors with clang

crowell:


src/JSystem/J3DGraphBase/J3DShape.cpp
```diff
@@ -2,7 +2,9 @@
 
 #include "JSystem/J3DGraphBase/J3DShape.h"
 
+#if !__APPLE__
 #include <malloc.h>
+#endif
```
PJB3005: This header can be removed actually, I think.

TakaRikka (approved):

PR #28, opened 2026-02-27
Isolate JKRHeap operator overloads

### LINKED ISSUE #25: Isolate game allocators (probably with a separate shared library)
JSystem relies heavily on overriding `operator new()` globally. This is a problem, as all our statically linked dependencies *also* start using JSystem's `operator new`. This includes all the complicated Aurora dependencies like Dawn. This has already caused problems (#19) and likely will continue to do so in the future. And no, it is not possible to "just remove the overloads and use the system allocator". Quoting myself:

> Okay, great, I need to isolate the actual TP code into its own dll. *sigh*
> 
> Specifically:
> 1. They override `void* operator new` globally
> 2. They override `void operator delete` but without a heap parameter so it has to *search through all heaps* to find the right one to free from
> 3. They seem to rely on being able to call `->freeAll()` to clear specific heaps like an arena allocator
> 
> These requirement seem impossible to reconcile without nuking `operator new` performance in the process (for eg other libraries like Dawn)
> 
> I'm thinking isolating the TP code to its own DLL to isolate the `void* operator new`... which is gonna be annoying but it'll have to do

There's like two ways I can think of fixing this:
1. Make the `operator new()` implementations `inline`, so they do not leave TUs that use them. This requires us to ensure with 100% confidence that the relevant header is included in *every* file that needs it, and not included where we *don't* need it.
2. Make all game code relying on JSystem compiled into its own shared library (dll, so, dylib) and have Aurora and all other dependencies live in a different shared library. Likely best approach is to have a `dusk.exe`, `game.dll` and `dusk_support.dll`? 

I think I prefer (2) as it's much saner to confirm we did it it right than (1).

PJB3005:
Fixes #25

This isolates the JKRHeap operator new/delete overloads. Every single new/delete site in the code has been replaced with a macro.

Sadly for new[] and delete[] we have to keep global operators. The global new[] just allocates into malloc() however, and delete[] goes into free() if it's not in a JKRHeap. So that's fine.


TakaRikka (commented):
overall looks fine, but for the powerpc eabi files and nw4hbm stuff can probably be left out since they don't use the JSystem overloads. the eabi is probably useless for pc anyway, and the nw4hbm is wii version specific

PJB3005:
> overall looks fine, but for the powerpc eabi files and nw4hbm stuff can probably be left out since they don't use the JSystem overloads. the eabi is probably useless for pc anyway, and the nw4hbm is wii version specific

oh yeah I should probably go through and clean those up huh

PR #34, opened 2026-03-01
Fix MSVC linker error for dBgW::CaptPoly

Lurs:
See commit


PJB3005:
image
uh, isn't the real problem that there are two versions of this file?

Lurs:
Oh, you are right, didn't check this. So this would probably be better fixed upstream in TP?

PR #36, opened 2026-03-05
fix visibility, turn off overridden new/delete when we call into aurora

crowell:
also fix a small memory stomper for good measure


PJB3005:
is there much of a point to this once #28 is merged?

crowell:
yes, we can't export the new/delete symbols still. the saveheap/setcurrentheap might be able to be removed, but not 100% sure. so at least part of this needs to stay.

PJB3005:
> yes, we can't export the new/delete symbols still.

But we don't anymore, in that PR?

crowell:
that branch doesn't build for me at all, so I can't verify/test.

PR #40, opened 2026-03-07
Fix build on main for mac/linux

crowell:
fixed up aurora linkage as well.


crowell:
someone on windows please check it still builds for them, because I can't make any promises with how GXColor was handled :)

PR #83, opened 2026-03-13
fix clang error on pointer narrowing

crowell:


src/d/actor/d_a_npc_tkj2.cpp
```diff
@@ -103,7 +103,7 @@ static int nodeCallBack(J3DJoint* i_joint, int param_2) {
         int jntNo = joint->getJntNo();
         J3DModel* model = j3dSys.getModel();
         npc_tkj2_class* i_this = (npc_tkj2_class*)model->getUserArea();
-        int userArea = (int)i_this;
+        int userArea = (int)(uintptr_t)i_this;
```
TakaRikka: this is prob more of a decomp error, but this var should prob be `fopAc_ac_c*` and should fix the issue the same

crowell: yeah, this is weird, userArea is just never used after? I'll change it to a `fopAc_ac_c *` though

TakaRikka: yeah its a normal pattern we see a lot where the devs had a separate base actor pointer that they used for accessing base actor members, even if it went unused or they removed w/e code was using it originally

PR #91, opened 2026-03-13
fix flowers

qwertyquerty:


include/d/actor/d_flower.h
```diff
@@ -103,6 +103,11 @@ class dFlower_packet_c : public J3DPacket {
     /* 0x12A48 */ u32 m_Jhana01DL_size;
     /* 0x12A4C */ u8* mp_Jhana01_cDL;
     /* 0x12A50 */ u32 m_Jhana01_cDL_size;
+
+#if TARGET_PC
+    GXTexObj mTexObj_l_J_Ohana00_64TEX{};
```
encounter, line 108: Use TGXTexObj which gives us automatic zero-initialization and destruction (no GXDestroyTexObj needed)

PR #93, opened 2026-03-15
JAUSoundTable-related crash fix

Irastris:
Allows loading into more areas, tested using the Lanayru Fishing Pond. Does not resolve the underlying issue of why these things are NULL, so that should be looked into further.

### DIFF
--- a/libs/JSystem/src/JAudio2/JAUSoundTable.cpp
+++ b/libs/JSystem/src/JAudio2/JAUSoundTable.cpp
@@ -15,7 +15,7 @@ void JAUSoundTable::init(void const* param_0) {
 }
 
 u8 JAUSoundTable::getTypeID(JAISoundID param_0) const {
-#if TARGET_PC
+#if DUSK_AUDIO_DISABLED
     if (this == NULL) {
         return 0xff;
     }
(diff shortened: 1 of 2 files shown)


TakaRikka:
should this go to pjb's branch?

Irastris:
Possibly, though since it does not actually implement any audio and just prevents the crash from that stuff not being right yet, I figured it'd be main-branch worthy.

Happy to pull into pjb-audio instead if you and @PJB3005 prefer.

PJB3005 (changes requested):
You can check for `DUSK_AUDIO_DISABLED` instead of `TARGET_PC` and then it definitely won't interfere with my branch

Irastris:
Made the requested change. Seems especially main-branch worthy since, even when it utilized TARGET_PC, it's nothing more than a few null checks so that later code that would crash is skipped. When audio support comes to main, these checks should become benign.

PJB3005 (approved):
You can check for `DUSK_AUDIO_DISABLED` instead of `TARGET_PC` and then it definitely won't interfere with my branch

PJB3005:
> When audio support comes to main, these checks should become benign.

My worry is more that they may hide a future bug's real root cause.

Irastris:
> My worry is more that they may hide a future bug's real root cause.

Fair enough, I hadn't considered that. Limiting them to when an audio implementation is disabled makes perfect sense then.

PJB3005:
> You can check for `DUSK_AUDIO_DISABLED` instead of `TARGET_PC` and then it definitely won't interfere with my branch

Why the fuck did GitHub re-post the previous review message. What?

PR #94, opened 2026-03-16
Bugfixes for some issues (cXyz in savegame, collision and fading)

### LINKED ISSUE #74: [Bug] Voiding out in city in the sky causes softlock on fade
Reproduce on file 2

https://discord.com/channels/1446645861529550869/1446648842387722411/1481807910228070561

### LINKED ISSUE #80: [BUG] All cXyz that exist inside of dSv_save_c or substructs need to be BE

### LINKED ISSUE #82: [Bug] integer div by 0 crsah in JUTFader that sometimes happens during fades
Image

Image

I got this one to happen by beating puppet zelda

Lurs:
Fixes #74
Fixes #80 
Fixes #82

### DIFF
--- a/include/d/d_com_inf_game.h
+++ b/include/d/d_com_inf_game.h
@@ -1084,7 +1084,11 @@ s32 dComIfGs_isDungeonItemWarp(int i_stageNo);
 void dComIfGs_BossLife_public_Set(s8);
 s8 dComIfGs_sense_type_change_Get();
 void dComIfGs_sense_type_change_Set(s8);
+#if TARGET_PC
+cXyz dComIfGs_getWarpPlayerPos();
+#else
 cXyz& dComIfGs_getWarpPlayerPos();
+#endif
 const char* dComIfGs_getWarpStageName();
 s16 dComIfGs_getWarpPlayerAngleY();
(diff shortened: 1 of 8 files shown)


Lurs:
Question is if I should keep AVOID_UB in libs/JSystem/src/JUtility/JUTFader.cpp or change it into TARGET_PC?

TakaRikka:
i think its prob fine as is as far as i understand

include/d/d_com_inf_game.h
```diff
@@ -1084,7 +1084,11 @@ s32 dComIfGs_isDungeonItemWarp(int i_stageNo);
 void dComIfGs_BossLife_public_Set(s8);
 s8 dComIfGs_sense_type_change_Get();
 void dComIfGs_sense_type_change_Set(s8);
+#if TARGET_PC
+cXyz dComIfGs_getWarpPlayerPos();
```
TakaRikka, line 1088: im curious if not returning references will cause logic issues

Lurs: I could not see any getPos().x = ... or getPos().set(...) patterns, so it should be ok. But I must admit that I did not fully understand the WarpPlayerPos concept of the game..

PR #98, opened 2026-03-17
Add tweak to enable/disable TV settings screen

caseif:
This adds a new toggle to the QoL menu to enable or disable the TV settings screen.

~~I'm not sure whether this is something that we want to be implement at this stage (or at all), so this PR also serves to get feedback on that.~~

~~I'm slightly iffy on how the new `levers` interface is currently set up - ideally I think we should have a distinct interface separate from ImGui stuff for accessing tweak settings, but it gets a little weird since ImGui works with pointers rather than callbacks. We could maintain separate globals for both the public tweaks interface and the UI layer and update the interface values when something in the UI changes, but that's a lot of added complexity and (IMO) unnecessary redundancy.~~

~~That all said, I do also anticipate that we'll eventually have a more sophisticated UI in place so I'm ultimately not too concerned about the weirdness right now.~~

### DIFF
--- a/include/d/d_s_name.h
+++ b/include/d/d_s_name.h
@@ -48,6 +48,7 @@ public:
     void FileSelectClose();
     void brightCheckOpen();
     void brightCheck();
+    void doPreLoadSetup();
     void changeGameScene();
 
     #if VERSION == VERSION_GCN_PAL
@@ -70,6 +71,9 @@ private:
     /* 0x41E */ u8 mWaitTimer;
     /* 0x41F */ u8 field_0x41f;
     /* 0x420 */ u8 field_0x420;
+#if TARGET_PC
+    bool mShowTvSettingsScreen;
+#endif
 };
 
 #endif /* D_S_D_S_NAME_H */
--- a/include/dusk/settings.h
+++ b/include/dusk/settings.h
@@ -27,6 +27,7 @@ struct UserSettings {
     struct {
         // QoL
         bool enableQuickTransform;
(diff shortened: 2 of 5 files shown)


TakaRikka (commented):
i feel like the levers stuff and adding a field to d_s_name might be unnecessary? what i would do is just have the var as part of the tweaks menu class, and then access it through the menu global var. this way everything is handled through the global imgui instance rather than a bunch of separate tweaks or globals

caseif:
My thinking is that it would be better to avoid coupling game code too closely with ImGui and instead have a shared interface between the two - that way if we move to a more user-friendly UI in the future (like the one in Ship of Harkinian, although idk which framework that uses) we can just hook it up to the common interface and not have to change any game code.

TakaRikka:
afaik ship is still using imgui, just styled heavily

caseif:
Ah okay, I didn't recognize it as ImGui so I assumed it was something different. In any event: my gut instinct is still to want to maintain a division between frontend and backend for the sake of SoC and tidiness, but if you don't think it's worth the added complexity then I can refactor this to have the field in the ImGui layer directly.

Also, I missed your point about the new field in `d_s_play` before - that only serves to cache a consistent value for the flag through the whole process since it has multiple stages, in case the user toggles it for some reason between selecting the save and the game starting. I could also move it to a static or something if you'd prefer so we're not touching the class structure, just as long as we're caching it somewhere.

caseif:
Rebased on latest `main` to utilize the new settings interface from #209.

PR #139, opened 2026-03-28
Make the movie player work

PJB3005:
THPs are *almost* just a bunch of JPEGs so the TL;DR is BE fixes and replacing the decoding with libjpeg-turbo.


src/d/actor/d_a_movie_player.cpp
```diff
@@ -4050,14 +4366,55 @@ static void daMP_ActivePlayer_Main() {
     }
 }
 
+#if TARGET_PC && 0
+#include "imgui.h"
+#endif
```
TakaRikka, line 4371: is this imgui stuff just temp debugging stuff? if so should prob be removed now or moved into the proper menu bar if still needed

PJB3005: It's behind `#if 0` so I figured leaving it in wouldn't hurt. I can remove it if you want though.

PR #144, opened 2026-03-29
Use separate heap for DVD decompression

caseif:
This fixes crashes when first entering North Faron cave and Renado's Sanctuary due to system heap exhaustion resulting from multiple threads allocating on the heap at once.

There's some sort of weird contention going on - for some reason the game gobbles up seemingly any amount of memory you throw at it when the two threads are competing for heap space. Using a separate heap in the async thread sidesteps the issue.


TakaRikka (commented):
this personally seems fine to me afaict, but would like someone more knowledgeable about this to look over it

YunataSavior (approved):

PJB3005 (approved):

PR #145, opened 2026-03-29
Use PCH to massively speed up compile time

PJB3005:
A change to a common header goes from 90s to 20s on my system. Crazy.


TakaRikka (commented):
not a big thing, but the pch could prob just go in `include` rather than `include/d`

PR #165, opened 2026-03-31
Disable suspicious assert causing cutscene crashes

PJB3005:
From my reading of the code this assert is likely incorrect. This throws on cases where usize has a sane-looking value (the length of the pContent null-terminated string), and from my understanding of how the data is parsed, this length is needed to figure out the location of the next "paragraph".


caseif:
According to debug this is supposed to be `uSize>0`. A similar issue existed in `object-particle` - I think it's probably a copy-paste error or something in the decomp. In any event I would just change it to that as a stopgap until the file can be cleaned up upstream.

PJB3005:
Yeah I don't have debug set up in decomp, so thanks for confirming.

PR #173, opened 2026-03-31
Prevent Epona minigame from crashing when Wolf Link (speedrun strat)

Irastris:
Clamps some array lookup values that originally weren't


TakaRikka (commented):
maybe should make these print some debug log that says its falling back? since this could hide some underlying issues like uninitialized vecs that could go unnoticed

Irastris:
Added the logging plus some of my (admittedly undercooked) insights. Maybe later I'll compile decomp with logging and see what happens to idx, var1 and var3 in Dolphin Emulator.

YunataSavior:
@Irastris can you watch the post Lanayru twilight Interlopers cutscene with this fix? In the original game, it crashes if you're wolf link

Irastris:
@YunataSavior Do you have a save to get me closer to that moment? I don't know what event flags and room layer would be necessary to do it manually

YunataSavior:
@Irastris you need to finish Lanayru twilight with the Master Sword obtained (early)

PR #193, opened 2026-04-01
Implement Quick Transform from Rando

PJB3005:


src/d/actor/d_a_alink_dusk.cpp
```diff
@@ -0,0 +1,74 @@
+#include "d/actor/d_a_alink.h"
```
TakaRikka, line 1: this probably doesn't need an entire file dedicated to it. i'd just put it next to the original procCoMetamorphose

src/f_ap/f_ap_game.cpp
```diff
@@ -721,6 +721,67 @@ void fapGm_After() {
     fopCamM_Management();
 }
 
+#if TARGET_PC
+// Analog L is currently not being used, so commented out
+// static f32 prevFrameAnalogL = 0.f;
+static f32 prevFrameAnalogR = 0.f;
+
+static bool DuskCheckButtonCombo(u32 combo, bool checkAnalog) {
```
TakaRikka: is there a reason to handle button checking separately with dusk rather than just use the ingame cpad functions?

TakaRikka (commented):
some notes and some conflicts that need fixing

TakaRikka:
i added an enhancements tab to handle features like this now. i'd like if this was added to the QoL section

PJB3005:
It's in its own file because my IDE absolutely *dies* dealing with the main link file. I can move it if really desired though.

src/f_ap/f_ap_game.cpp
PJB3005: Honestly I didn't really look at this to see if they were suitable.

TakaRikka: you should be able to do something basically just like 
```cpp
if (mDoCPd_c::getHoldR(PAD_1) && mDoCPd_c::getTrigY(PAD_1))
```
to check if you're holding R and pressed Y. ive handled it like this in mods before and it seemed to work just fine

MelonSpeedruns: Maybe make it:
```cpp
if ((mDoCPd_c::getHoldR(PAD_1) && mDoCPd_c::getTrigY(PAD_1)) || (mDoCPd_c::getTrigR(PAD_1) && mDoCPd_c::getHoldY(PAD_1)))
```
just in case someone does the opposite button combo.

TakaRikka:
> It's in its own file because my IDE absolutely _dies_ dealing with the main link file. I can move it if really desired though.

maybe something like `d_a_alink_custom.inc` or `d_a_alink_dusk.inc`? just so that if there's more similar needs in the future we dont need to create a bunch of files

PJB3005:
> > It's in its own file because my IDE absolutely _dies_ dealing with the main link file. I can move it if really desired though.
> 
> maybe something like `d_a_alink_custom.inc` or `d_a_alink_dusk.inc`? just so that if there's more similar needs in the future we dont need to create a bunch of files

Well it wouldn't be a `.inc` since the whole point is to have it be a separate C++ file, but sure.

TakaRikka:
> Well it wouldn't be a `.inc` since the whole point is to have it be a separate C++ file, but sure.

sure thats fine

PR #209, opened 2026-04-03
Refactor Dusk settings out to common interface used by game and UI code

caseif:
This moves the settings variables out from the ImGui TUs to a common interface which can be included independent of any UI code. I'm not super content with how everything is named/categorized atm, but I'm also not quite sure what it _should_ look like and we can always iterate on that as we add more settings.

This interface can also be used to expose functions for saving/loading persistent settings in the (hopefully near) future.

This PR also includes changes to move hotkey definitions to a separate header so that they're all defined in the same place, rather than wherever their respective menu items happen to be defined.

### DIFF
--- a/include/dusk/settings.h
+++ b/include/dusk/settings.h
@@ -1,0 +1,71 @@
+#ifndef DUSK_CONFIG_H
+#define DUSK_CONFIG_H
+
+namespace dusk {
+
+// Persistent user settings
+
+struct UserSettings {
+    // Program settings
+
+    struct {
+        // Video
+        bool enableFullscreen;
+    } video;
+
+    struct {
+        // Audio
+        float masterVolume;
+        float mainMusicVolume;
+        float subMusicVolume;
+        float soundEffectsVolume;
+        float fanfareVolume;
+    } audio;
+
+    // Game settings
+
+    struct {
+        // QoL
+        bool enableQuickTransform;
+
+        // Preferences
+        bool enableMirrorMode;
+        bool invertCameraXAxis;
+
+        // Graphics
+        bool enableBloom;
+        bool useWaterProjectionOffset;
+
+        // Cheats
+        bool enableFastIronBoots;
+
+        // Technical
+        bool restoreWiiGlitches;
+    } game;
+};
+
+UserSettings& getSettings();
+
+// Transient settings
+
+struct CollisionViewSettings {
+    bool enableTerrainView;
+    bool enableWireframe;
+    bool enableAtView;
+    bool enableTgView;
+    bool enableCoView;
+    float terrainViewOpacity;
+    float colliderViewOpacity;
+    float drawRange;
+};
+
+struct TransientSettings {
+    CollisionViewSettings collisionView;
+};
+
+TransientSettings& getTransientSettings();
+
+}
+
+#endif // DUSK_CONFIG_H
+
--- a/files.cmake
+++ b/files.cmake
@@ -1341,6 +1341,7 @@ set(DUSK_FILES
         src/dusk/extras.c
         src/dusk/extras.cpp
         src/dusk/globals.cpp
+        src/dusk/settings.cpp
         #src/dusk/m_Do_ext_dusk.cpp
         src/dusk/imgui/ImGuiConsole.hpp
         src/dusk/imgui/ImGuiConsole.cpp
--- a/include/dusk/hotkeys.h
+++ b/include/dusk/hotkeys.h
@@ -1,0 +1,19 @@
+#ifndef DUSK_HOTKEYS_H
+#define DUSK_HOTKEYS_H
+
+namespace dusk::hotkeys {
+
+constexpr const char* DO_RESET = "Ctrl+R";
+
+constexpr const char* TOGGLE_FULLSCREEN = "F11";
+
+constexpr const char* SHOW_PROCESS_MANAGEMENT = "F2";
+constexpr const char* SHOW_DEBUG_OVERLAY = "F3";
+constexpr const char* SHOW_HEAP_VIEWER = "F4";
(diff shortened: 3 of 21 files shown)


include/dusk/settings.h
```diff
@@ -0,0 +1,69 @@
...
+};
+
+struct CheatsSettings {
+    bool enableFastIronBoots;
+};
+
+extern GameplaySettings g_gameplay;
+extern GraphicsSettings g_graphics;
+extern TweaksSettings g_tweaks;
+extern CheatsSettings g_cheats;
```
TakaRikka: i think i'd prefer these all under an enhancements master struct, with tweaks and gameplay stuff combined into one group, and mirror mode moved back to graphics

Irastris: Not that I was asked but I feel like that would completely defeat the purpose of cleanup. Caseif's PR looks great to me. Also, I think Mirror Mode would make more sense under gameplay than it would tweaks or graphics.

caseif: I really think we should keep some sort of shallow organization rather than putting everything in a flat struct, given that the number of settings is likely to grow quite a bit. Maybe we could keep the existing structs but store them in a single global via a master struct?

With regard to combining the tweaks and gameplay categories, my reasoning is that the items currently under tweaks are unlikely to be of much interest to a typical player so it makes sense to keep them separate so as not to clutter the settings that will be more widely used. As an example, as a non-speedrunner I have zero interest in Wii 1.0 glitches outside of being a technical curiosity.

For mirror mode, I was a little iffy on this but I think it conceptually fits better under tweaks as 1) it's hard to consider it an "enhancement" per se, and 2) it goes way beyond being a purely graphical change - think of mirror courses in Mario Kart being a separate gamemode. Granted, the first point is more subjective and I'm maybe amenable to putting it under gameplay instead, but I definitely don't think it belongs under graphics.

TakaRikka: its mostly just that i dont like having a bunch of globals to manage, at least when its all in one super struct you only have one global to worry about. if theres a better solution i'd be interested

as for the categories, i came up with something like this while i was just brainstorming ideas the other day
```
  - Fixes/Restorations
    - Enable Wii 1.0 glitches
  - Graphics
    - Hide UI
    - Disable LOD?
    - Increase draw distance?
    - Mirror world (Wii mode)
  - Items
    - Bow Crosshair
    - Remove bomb limit?
  - QoL
    - Autosave?
    - Shops always open at any time of day
    - Fast text
    - Input buffering?
    - Climbing speed change
    - Fast Iron Boots
    - Fast chests
    - No Fishing Bottle RNG
    - More lenient Zelda/bo RNG?
    - No interrupting Midna's Lament
    - Display Time of Day?
  - Difficulty?
  - Cheats
  - Cosmetics Editor
  - Speedrun mode?
```
which may help explain my thought process on the categories

caseif: I think that list generally makes sense, although I still think that mirror is better placed somewhere other than graphics given that it affects the actual gameplay too and isn't implemented as just a graphical effect like the other options are. I would also lean towards putting a lot of the QoL items (shops open, climb speed, fast boots, RNG changes) under Cheats since they arguably make the game easier, although for things like shops being open 24 hours it's maybe a little less clear-cut. The Items toggles are also a little suspect IMO - I think we should limit the QoL label to tweaks that don't alter gameplay in a meaningful way.

Also, I definitely don't think a flag to enable glitches belongs under Fixes. Restorations could fit, although I'm not sure how many other options would actually end up there.

TakaRikka: i suppose moving mirror mode would be fine. climb speed or any other changes from the HD version i def think are QoL. fast boots as a cheat is fair. shops/rng are debatable, can come back to that later. the item ones would be a separate thing from QoL. some of these ideas i got from ship fwiw, so i had influence from them

yea the glitches one i wasnt sure exactly where it would fit best. personally i think its really cool to enable older behavior from 1.0, but dont know exactly how to categorize it

caseif: Ah, I haven't played TPHD so I didn't realize climb speed was changed there - in that case I agree we should categorize anything from HD as QoL. I do see the argument for shops/RNG being QoL but yeah, we can cross that bridge when we get to it. Also my bad with the Items flags, I got some wires crossed while reading that.

For the 1.0 glitches flag, maybe "Technical" could make sense as a category? We can also always shuffle it around if more flags in that vein end up being added.

For the invert camera X axis flag, what do you think we should do with that? I feel like a QoL category makes Gameplay as a category somewhat redundant, but I'm also not sure it's a QoL tweak so much as a preference. Maybe we could have a Tweaks or Preferences category for that and mirror mode? I'm not sure what would make the most sense, though - maybe Gameplay would still be more appropriate.

caseif: Actually, in shuffling things around I might be coming around to having everything in a flat struct and just using comments for organization. If we're going to be potentially moving things as development progresses we're going to either need to also update our model constantly, or have our model and UI drift out of sync which would be confusing.

@Irastris Any thoughts on this?

Irastris: Comments doesn't sound bad so if that's something you and Taka were able to agree on, have at it

caseif:
@TakaRikka Let me know how the latest changes here look.

TakaRikka:
i'd still be interested in sub-structing some of the groups like having video/audio/enhancements anonymous substructs so that when you access them it's like `dusk::getSettings().enhancements.mirrorMode` or `dusk::getSettings().audio.masterVolume`. but if thats too much for right now, can come back to it later

caseif:
@TakaRikka That's definitely reasonable given video/audio settings are pretty unambiguous. I've moved everything into anonymous structs, although I used `game` as the name for the enhancement struct for the sake of brevity.

PR #216, opened 2026-04-04
Non-Stop Midna's Lament Audio Enhancement

MelonSpeedruns:
Prevents enemy music while Midna's Lament is playing.

### DIFF
--- a/src/Z2AudioLib/Z2SoundObjMgr.cpp
+++ b/src/Z2AudioLib/Z2SoundObjMgr.cpp
@@ -108,7 +108,9 @@ void Z2SoundObjMgr::searchEnemy() {
     twilightBattle_ = 0;
 
     #if TARGET_PC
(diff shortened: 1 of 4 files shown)


src/Z2AudioLib/Z2SoundObjMgr.cpp
```diff
@@ -107,6 +107,13 @@ static Z2EnemyInfo mEnemyInfo[64] = {
 void Z2SoundObjMgr::searchEnemy() {
     twilightBattle_ = 0;
 
+    #if TARGET_PC
+    if (Z2GetSeqMgr()->checkBgmIDPlaying(Z2BGM_MIDNA_SOS)) {
```
PJB3005: Forgot to actually check the option is enabled?

PJB3005 (changes requested):

src/Z2AudioLib/Z2SoundObjMgr.cpp
MelonSpeedruns: Fixed!

PR #219, opened 2026-04-04
No Low HP Sound Enhancement

MelonSpeedruns:
Disable the beeping sound when having low health.


TakaRikka:
lgtm after conflicts are fixed

PR #221, opened 2026-04-04
Config system v1

### LINKED ISSUE #200: Configuration/settings system

PJB3005:
Roughly inspired by what I've learned from my work on Space Station 14, without some of the unnecessary cruft and complexity.

Implementation is relatively simple once I figured out all the template order shenanigans.

Fixes #200


TakaRikka:
i dont really have any problems with this in theory, but would like to know what @encounter and @caseif think

src/dusk/config.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+using json = nlohmann::json;
+
+aurora::Module DuskConfigLog("dusk::config");
+
+static absl::flat_hash_map<std::string_view, ConfigVarBase*> RegisteredConfigVars;
+static bool RegistrationDone;
+
+static std::string GetConfigJsonPath() {
+    dusk::ScopeValue folder(SDL_GetPrefPath(dusk::OrgName, dusk::AppName), SDL_free);
```
encounter: We should use the configPath returned in AuroraInfo

caseif:
Skimming over this nothing really stands out to me, other than it should be refactored on top of the changes from #209 which I think should be pretty straightforward, and maybe the new headers could be merged into a single `config.hpp` (except `io.hpp` which I think makes sense in its own file). I'll take a closer look tomorrow when I'm less sleepy, but overall it looks mostly good to me.

src/dusk/config.cpp
PJB3005: This isn't possible because we need to load config *before* Aurora is initialized (e.g. to load backend config). I'll pass the config dir into Aurora explicitly though.

PJB3005:
> and maybe the new headers could be merged into a single `config.hpp`

The current header split (`config.hpp` vs `config_var.hpp`) is explicitly to avoid causing serious compile time issues due to including all the implementation details. Also makes iteration less of a pain. As for the rest, `appname.hpp` seems reasonable to be separate since it's not strictly tied to the config system.

caseif:
I'm not sure I follow - shouldn't there be no difference practically speaking since `config.hpp` already pulls in `config_var.hpp`?

For `appname`, that's fair although I think the name is maybe a little too specific - something broader like `constants` or `app_info`/`app_meta` would be better IMO.

PJB3005:
> shouldn't there be no difference practically speaking since `config.hpp` already pulls in `config_var.hpp`?

Files like `settings.h` are imported into basically every file of game code. It only includes `config_var.hpp`, not `config.hpp`.

caseif:
I think I might still be missing something - what about `config.hpp` is problematic when including it? It looks like it contains pretty minimal implementation code, and the number of template instantiations should also be fairly small given the limited number of data types in use.

PJB3005:
> I think I might still be missing something - what about `config.hpp` is problematic when including it? It looks like it contains pretty minimal implementation code, and the number of template instantiations should also be fairly small given the limited number of data types in use.

It includes the JSON lib

caseif:
Ah okay, that makes sense! Could you just add a note to the top of both files to make that clear explicitly?

src/m_Do/m_Do_main.cpp
```diff
@@ -233,10 +237,51 @@ static void aurora_imgui_init_callback(const AuroraWindowSize* size) {
...
+        try {
+            cVar->getImpl()->loadFromArg(*cVar, value);
+        } catch (const std::exception& e) {
+            DuskLog.fatal("Unable to parse: '{}': {}", value, e.what());
+        }
+    }
+}
+
+static const char* CalculateConfigPath() {
+    const auto result = SDL_GetPrefPath(dusk::OrgName, dusk::AppName);
```
encounter, line 270: We can actually pass nullptr for the org name, unless we really want it to be %appdata%\TwilitRealm\Dusk instead of %appdata%\Dusk

PJB3005: > We can actually pass nullptr for the org name, unless we really want it to be %appdata%\TwilitRealm\Dusk instead of %appdata%\Dusk

```
 * Due to historical mistakes, `org` is allowed to be NULL or "". In such
 * cases, SDL will omit the org subdirectory, including on platforms where it
 * shouldn't, and including on platforms where this would make your app fail
 * certification for an app store. New apps should definitely specify a real
 * string for `org`.
```

PJB3005:
> Ah okay, that makes sense! Could you just add a note to the top of both files to make that clear explicitly?

Done.

src/m_Do/m_Do_main.cpp
encounter: oh, rip

PR #231, opened 2026-04-05
Armors Quick-Toggle

Captainkittyca2:
After unlocking the armors, you can toggle between them by pressing d-pad down once or twice.

NOTE:
For testing purposes, I made it so enabling the enhancement automatically give you the armors if you don't have them so people can test the feature. We can remove that if statement before public release or after developers test Armors Quick-Toggle.

Here's how it works:

Green/Magic -> Zora OR Zora -> Green: Press D-Pad down once
Green/Zora -> Magic OR Magic -> Green: Press D-Pad Down twice

Also, when enabling the enhancement, you can't open item wheel with d-pad down, only d-pad up. When disabling, you can open item wheel with d-pad down.

Finally, if you guys would prefer pressing another button for Armors Quick-Toggle, we can discuss that too.


Captainkittyca2:
btw, line 18 to 20 in d_a_alink_dusk.cpp is the temporary code that I mentioned in the NOTE above. Also the purpose of lines 53 to 59 is cooldown period, otherwise game will crash if you change armors right after immediately already changing armors

src/d/d_menu_window.cpp
```diff
@@ -126,6 +126,9 @@ BOOL dMw_UP_TRIGGER() {
 }
 
 BOOL dMw_DOWN_TRIGGER() {
+    #if TARGET_PC
+    if (dusk::getSettings().game.enableArmorsQuickToggle) return NULL;
+    #endif
     return mDoCPd_c::getTrigDown(PAD_1) != 0;
```
TakaRikka: idk how i feel about this, this feels a little too hacky. also its returning a bool, so it shouldnt be returning NULL

src/d/actor/d_a_alink_dusk.cpp
```diff
@@ -3,6 +3,62 @@
...
 #include "d/d_meter2_draw.h"
 #include "d/d_meter2_info.h"
+#include "m_Do/m_Do_audio.h"
+
+bool armorQuickToggleVar1 = false;
+bool armorQuickToggleVar2 = false;
+uint32_t countyer = 0;
+uint32_t armorQuickToggleTimer1 = 0;
+uint32_t armorQuickToggleTimer2 = 0;
+
```
TakaRikka: u32, and more appropriate names for these vars would be preferred

src/d/actor/d_a_alink_dusk.cpp
```diff
@@ -3,6 +3,62 @@
...
+void daAlink_c::handleArmorsQuickToggle() {
+    if (!dusk::getSettings().game.enableArmorsQuickToggle) return;
+
+    // temporary if statement code for automatically having magic and zora armors unlocked to test the mod
+    if (!dComIfGs_isItemFirstBit(dItemNo_WEAR_ZORA_e)) {
+        dComIfGs_onItemFirstBit(dItemNo_WEAR_ZORA_e); dComIfGs_onItemFirstBit(dItemNo_ARMOR_e);
```
TakaRikka: formatting is kinda all over the place, run clang format or manually reformat to the style generally seem elsewhere

TakaRikka (commented):
some cleanup needed. also im not really sold on the controls, disabling dpad down feels hacky and needing multiple presses to do different actions feels unnatural. also considering you dont really need to change armors very often in this game compared to something like boots in oot makes this feel a little unnecessary. maybe with some more polish to make it feel like a natural game mechanic i could have my opinion changed

src/d/actor/d_a_alink_dusk.cpp
Captainkittyca2: ah that line of code is only for testing purposes so devs can automatically have armors unlocked, I'll delete lines 18 to 20 it after people test the feature

src/d/actor/d_a_alink_dusk.cpp
Captainkittyca2: ah ok, i will change it

src/d/d_menu_window.cpp
Captainkittyca2: ah should I return false? I'm actually not sure tbh since I always do return NULL in my mods for this function. I'll try return false.

Also, Idk if it's really hacky. I think disableRupeeCutscenes does something similar in line 372 in d_s_name.cpp

TakaRikka: NULL is for pointers, this isnt returning a pointer. also thats a different case, this is completely disabling a button. disabling rupee cutscenes just bypasses some logic for unsetting flags

Captainkittyca2: ah I see.
I'll make the changes to false and the u32 isntead of uint32_t and will notify back

Captainkittyca2:
> some cleanup needed. also im not really sold on the controls, disabling dpad down feels hacky and needing multiple presses to do different actions feels unnatural. also considering you dont really need to change armors very often in this game compared to something like boots in oot makes this feel a little unnecessary. maybe with some more polish to make it feel like a natural game mechanic i could have my opinion changed

Can you check latest formatting and other new changes I made in the PR. I think I addressed most, if not all, problems. Formatting is better, variable names is more accurate to what they do, It's now R+X rather than D-Pad Down

Captainkittyca2:
also, I removed the temporary line of code in d_a_alink_dusk.cpp that gives you magic and zora armors qutomatically if you don't have them. So now would need to unlock them to use it (you can edit the code locally to give urself magic/zora armor for testing purposes)

src/d/d_menu_window.cpp
Captainkittyca2: So I removed any modifications to that function in d_menu_window, since Armors Quick-Toggle now uses R+X

src/d/actor/d_a_alink_dusk.cpp
Captainkittyca2: I think formatting now is fixed but I don't want to click "Resolve conversation" until other check, in case there's some other issues that I may have not noticed.

src/d/actor/d_a_alink_dusk.cpp
Captainkittyca2: changed to u32 and changed the var names to more accurate names

Captainkittyca2:
I updated it to be D-Pad Left for Zora/Green and D-Pad Right for Magic/Green.

src/d/actor/d_a_alink_dusk.cpp
Captainkittyca2: @TakaRikka Ok I just found out how to do the clang format thing. I never used clang-format command before (at least not directly) so I wasn't sure how to do it, but I recently looked it up and learned how to use it and ran it on handleArmorsQuickToggle. Now formatting should be good to go.

PR #237, opened 2026-04-05
Feature/dynamic aspect ratio

qwertyquerty:


caseif:
Could this maybe be rebased on main before merging to get rid of all the merge commits?

Also, we may want to gate this behind a settings flag since it still has quite a few rough edges.

PJB3005:
> Could this maybe be rebased on main before merging to get rid of all the merge commits?

That sounds extremely difficult due to the nature of rebasing. I'd really rather leave it as-is.

> Also, we may want to gate this behind a settings flag since it still has quite a few rough edges.

Will be handled by https://github.com/TwilitRealm/dusk/pull/238

caseif:
I can give it a go and push it to a separate branch.

MelonSpeedruns:
I have merged main into this branch, it's ready to be pushed into main.

PR #239, opened 2026-04-05
use "Cmd" instead of "Ctrl" for hotkeys on Apple

crowell:


MelonSpeedruns:
LGTM, though what about F11?

crowell:
macOS has a global fullscreen/not full screen cmd+ctrl+f, but I'm getting GPU errors when using it :\

MelonSpeedruns:
Hmmm might want to fix that as part of this PR then?

PJB3005 (approved):

PR #240, opened 2026-04-06
First run config presets

qwertyquerty:
image


include/dusk/config.hpp
```diff
@@ -111,6 +111,13 @@ void Save();
  */
 ConfigVarBase* GetConfigVar(std::string_view name);
 
+/**
+ * \brief Returns true if no config file was found on the last LoadFromUserPreferences() call.
+ *
+ * For detect first run to prompt the user to choose a preset.
+ */
+bool WasConfigFileMissing();
```
crowell: this is a weird name for the function, maybe "IsFirstRun" is better?

src/dusk/config.cpp
```diff
@@ -187,6 +188,7 @@ void dusk::config::LoadFromFileName(const char* path) {
     } catch (const std::system_error& e) {
         if (e.code() == std::errc::no_such_file_or_directory) {
             DuskConfigLog.info("Config file did not exist, staying with defaults");
+            s_configFileMissing = true;
```
PJB3005: Instead of doing this, I'd just make a CVar that's "was preset chosen" and set it to true afterwards.

src/dusk/imgui/ImGuiFirstRunPreset.cpp
```diff
@@ -0,0 +1,129 @@
...
+    s.game.hideTvSettingsScreen.setValue(true);
+    s.game.noReturnRupees.setValue(true);
+    s.game.disableRupeeCutscenes.setValue(true);
+    s.game.noSwordRecoil.setValue(true);
+    s.game.fastClimbing.setValue(true);
+    s.game.noMissClimbing.setValue(true);
+    s.game.fastTears.setValue(true);
+    s.game.noLowHpSound.setValue(true);
+    s.game.midnasLamentNonStop.setValue(true);
+    s.game.enableFastIronBoots.setValue(true);
```
caseif: This is a pretty substantial gameplay change - I don't think we should include it in a preset.

caseif:
I feel like it might make sense to also include HD features like fast climbing in the default preset.

src/dusk/imgui/ImGuiFirstRunPreset.cpp
MelonSpeedruns: I do agree with iron boots and the low hp sound.

TakaRikka:
default should be similar to HD, vanilla is still vanilla, while QoL should be called something like "enhanced" or something and have HD features + the extra QoL stuff.

src/dusk/imgui/ImGuiFirstRunPreset.cpp
caseif: I disagree on the low HP sound, especially if we're recommending it for experienced players. Alternatively, maybe we could implement a 3 or so second fade-out instead of removing it outright?

caseif:
Also, I think it would be good to have a menu item to allow selecting a preset at any point after the first run.

PR #241, opened 2026-04-06
Double heap sizes for various heaps (Midna eye)

Lurs:
Scale various hardcoded heap estimates 2x on TARGET_PC to account for 64 bit. Fixes Midna's eye. Fixes also various JKRHeap overflow fallbacks and a BE index table bug in setupBBoardInfo. ~~I feel like there are other places where heaps could be too small and I have tested not so much, so only Draft here~~. Also doubled the room heaps for `d_a_bg Anms`.

### DIFF
--- a/src/d/d_k_wmark.cpp
+++ b/src/d/d_k_wmark.cpp
@@ -6,6 +6,7 @@
 #include "d/dolzel.h" // IWYU pragma: keep
 
 #include "d/d_k_wmark.h"
+#include "dusk/memory.h"
 #include "JSystem/J3DGraphBase/J3DMaterial.h"
 #include "SSystem/SComponent/c_math.h"
 #include "d/actor/d_a_player.h"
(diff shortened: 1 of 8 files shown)


src/d/d_k_wmark.cpp
```diff
@@ -33,7 +33,11 @@ int dkWmark_c::create() {
         mColorType = this->parameters;
     }
 
+#if TARGET_PC
+    mpHeap = mDoExt_createSolidHeapFromGameToCurrent(0x1100, 0x20);
+#else
     mpHeap = mDoExt_createSolidHeapFromGameToCurrent(0x880, 0x20);
+#endif
```
PJB3005: Can use the `HEAP_SIZE()` macro for this.

src/d/d_kankyo.cpp
```diff
@@ -1175,7 +1175,11 @@ static void undwater_init() {
     J3DModelData* modelData2 = (J3DModelData*)dComIfG_getObjectRes("Always", 0x1D);
     JUT_ASSERT(1867, modelData2 != NULL);
 
+#if TARGET_PC
+    g_env_light.undwater_ef_heap = mDoExt_createSolidHeapFromGameToCurrent(0xC00, 0x20);
+#else
     g_env_light.undwater_ef_heap = mDoExt_createSolidHeapFromGameToCurrent(0x600, 0x20);
+#endif
```
PJB3005: Ditto.

Lurs: Ok

PJB3005 (approved):

PR #243, opened 2026-04-06
Runtime rel/dol asset loader

qwertyquerty:
Relies on https://github.com/encounter/aurora/pull/99


src/dusk/dvd_asset.cpp
```diff
@@ -0,0 +1,106 @@
...
+    
+    s_dolData = p;
+    s_dolSize = sz;
+
+    const BE(u32)* hdr = (const BE(u32)*)s_dolData;
+    s_dolSectionCount = 0;
+
+    // 0x00: text file offsets 0x12: text vaddrs 0x24: text sizes
+    for (int i = 0; i < 7;  i++) {
+        u32 off = hdr[0x00+i], addr = hdr[0x12+i], sz = hdr[0x24+i];
```
crowell: can you do this with a struct?

src/dusk/dvd_asset.cpp
```diff
@@ -0,0 +1,106 @@
...
+        if (sz > 0 && off > 0) {
+            s_dolSections[s_dolSectionCount++] = {off, addr, sz};
+        }
+    }
+    // 0x07: data file offsets 0x19: data vaddrs 0x2B: data sizes
+    for (int i = 0; i < 11; i++) {
+        u32 off = hdr[0x07+i], addr = hdr[0x19+i], sz = hdr[0x2B+i];
```
crowell: same, do this with a struct

src/d/actor/d_flower.inc
```diff
@@ -104,6 +110,16 @@ static u8 l_flowerTexCoord[] = {
     0x3E, 0xA7, 0x67, 0x4D, 0x3C, 0x14, 0x46, 0x74, 0x3E, 0xA7, 0x73, 0xE2, 0xBC, 0x2F, 0x46, 0xAA,
     0x3E, 0xA7, 0x72, 0xD6, 0xBD, 0x2F, 0x46, 0xAA};
 
+#if TARGET_PC
+static u8* l_J_hana00DL_get()   { static u8 buf[0x150]; static bool _ = (dusk::LoadArchivedRelAsset(buf, 'AMEM', "d_a_grass.rel", 0x9D20, 0x150), true); return buf; }
+static u8* l_J_hana00_cDL_get() { static u8 buf[0xDE];  static bool _ = (dusk::LoadArchivedRelAsset(buf, 'AMEM', "d_a_grass.rel", 0x9E80, 0xDE),  true); return buf; }
```
crowell, line 115: alignment is broken with these? is that important?

eg: 
```
unsigned char l_J_hana00_cDL[] ATTRIBUTE_ALIGN(32)
```
had explicit alignment. Maybe important, maybe not.

src/dusk/dvd_asset.cpp
qwertyquerty: done

src/dusk/dvd_asset.cpp
qwertyquerty: done

src/d/actor/d_flower.inc
PJB3005: It's not. Those attributes aren't even respected on MSVC.

TakaRikka: dl's being aligned to 32 was important for original hardware. i dont think it matters in aurora though afaik

src/d/actor/d_a_player.cpp
```diff
@@ -368,7 +367,13 @@ JKRHeap* daPy_anmHeap_c::setAnimeHeap() {
 }
 
 #if !PLATFORM_WII
+#if TARGET_PC
+#include "dusk/dvd_asset.hpp"
+static const u8* l_sightDL_get() { static u8 buf[0x89]; static bool _ = (dusk::LoadDolAsset(buf, 0x803BA0C0, 0x89), true); return buf; }
```
TakaRikka: maybe want to add a platform check for these ones? cause theoretically jp/pal should already pretty much work with little issue, but if we're checking hard-coded addresses then it'll break those versions

PR #251, opened 2026-04-06
Improve readability of hideTvSettingsScreen option + Don't play 2 sounds when selecting a save

MelonSpeedruns:


caseif:
The value was being cached to prevent any weirdness in case the user toggled the flag while loading a save since it's used across multiple transitions; I don't think that should be changed.

MelonSpeedruns:
> The value was being cached to prevent any weirdness in case the user toggled the flag while loading a save since it's used across multiple transitions; I don't think that should be changed.

Reverted the flag and made the sfx change use it!

src/d/d_file_select.cpp
```diff
@@ -1193,7 +1189,13 @@ void dFile_select_c::menuSelect() {
 
 // Handles copy / start / delete actions depending on which menu is selected from menuSelect
 void dFile_select_c::menuSelectStart() {
+    #if TARGET_PC && !SHOW_TV_SETTINGS_SCREEN
```
PJB3005: Uh, this resolves to `(this->mShowTvSettingsScreen)`, which doesn't make sense in a preprocessor context.

PJB3005 (changes requested):

src/d/d_file_select.cpp
MelonSpeedruns: Will fix tomorrow!

caseif: Good catch, this should be removed and the below if condition changed to `SHOW_TV_SETTINGS_SCREEN || mSelectMenuNum != 1` (I think).

PR #284, opened 2026-04-08
Sync up the Ordon Spring kidnapping cutscene music

MelonSpeedruns:
Dusk runs too fast compared to gamecube, and the music doesn't have time to finish since the video is super early.

This lowers the framerate during the cutscene to make the music work in sync with the video.

This whole problem is similar to Majora's Mask, where the cutscene with the 4 giants is faster on the Wii U version, and the music doesn't have time to end.

Ideally this is only temporary until we can figure out how to edit cutscenes and shorten some timings.

### DIFF
--- a/src/d/actor/d_a_demo00.cpp
+++ b/src/d/actor/d_a_demo00.cpp
@@ -85,6 +85,12 @@ static void dummy() {
 }
 
 inline int daDemo00_c::create() {
+    #if TARGET_PC
+    if (g_dComIfG_gameInfo.play.getEvent()->mEventId == 0x903) {
+        mDoGph_gInf_c::setTickRate(OS_TIMER_CLOCK / 27);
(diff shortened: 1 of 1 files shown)


TakaRikka:
yea this seems pretty hacky, but i dont know what the alternative is. maybe add a note at least about it? also will this break with interpolation?

MelonSpeedruns:
Let me try interpolation right away! I'll edit my comment here with the results.

Irastris:
The interpolation branch is currently hardcoded to 30hz for logic, it does not respect setTickRate() changes.

MelonSpeedruns:
After discussing, the fix would be for end-users to provide their own optional Wii rom, and extract the file `ToalVillage_d03.ast` into Appdata, and read it as an external mod file.

Issue #282, opened 2026-04-08
Crash at the end of the Armogohma cutscene


crowell:
macOS/clang RelWithDebInfo build

repro: 
use map loader to warp to "Temple of Time" -> "Armogohma Area"

let cutscene play

```
(lldb) bt
* thread #20, name = 'SDLAudioP15', stop reason = EXC_BAD_ACCESS (code=2, address=0x16e528000)
  * frame #0: 0x0000000102e87074 dusk`JASSeqParser::cmdPrintf(this=<unavailable>, param_0=0x00000005c12512c8, param_1=<unavailable>) at JASSeqParser.cpp:868:21 [opt]
(lldb) fr v
(JASSeqParser *) this = <could not evaluate DW_OP_entry_value: no parent frame with reg ctx>

(JASTrack *) param_0 = 0x00000005c12512c8
(u32 *) param_1 = <no location, value may have been optimized out>

(u8[4]) stack_c = "\U00000001"
(u32[4]) stack_10 = ([0] = 216, [1] = 1, [2] = 0, [3] = 0)
(char[128]) buffer = "_START\0\0\0\0\x80\xbf\0\0\0\0\0\0\x80?\0\0\0\0\xa0\xe2\b\U00000004\U00000001\0\0\0\0\xd0\b\U00000004\U00000001\0\0\0H'\x89\U00000003\U00000001\0\0\0hŏ\U00000003\U00000001\0\0\0H\U0000001bRn\U00000001\0\0\0\0\0\0\0\0\0\0\0\U00000004\0\0\0\0\0\0\0x\a\x89\U00000003\U00000001\0\0\0\xa8\x87#\xc1\U00000005\0\0\0\U00000014\0\0\0\0\0\0\0 \U0000001bRn\U00000001\0\0\0hv\xe8\U00000002\U00000001\0\0\0x\U0000001bRn\U00000001"
(JASSeqCtrl *) seqCtrl = 0x00000005c12512c8
(u32) r30 = 0
(u32) i = 6458
```

check the value of `i` and `r30` there must be some memory corruption around here or something...

crowell:
does not repro on Debug builds, so maybe it's an optimization of UB?

PR #286, opened 2026-04-08
Fix Crash at the end of the Armogohma cutscene #282

crowell:
zero stack buffers in JASSeqParser::cmdPrintf on PC builds


YunataSavior:
This should probably utilize the `AVOID_UB` define, no?

crowell:
changed to use `AVOID_UB`, ideally we might also just want `-ftrivial-auto-var-init=zero` as well.

PR #288, opened 2026-04-08
Free Magic Armor

SuperDude88:
Adds a cheat to remove the rupee drain for using the magic armor. Currently blocks all damage like the original armor, but it could be changed to be purely cosmetic/double defense/only consume rupees on damage if one of those are a more interesting option.


MelonSpeedruns:
I think we could simply make it the TPHD magic armor. I think it was changed to only remove rupees when you get damaged? Can't remember...

EDIT: Nvm! I'm thinking of wind waker hd!

PR #296, opened 2026-04-09
Add Tracy stuff

PJB3005:


include/d/d_com_inf_game.h
```diff
@@ -3109,114 +3111,133 @@ inline void dComIfGp_particle_calcMenu() {
 }
 
 inline void dComIfGp_particle_draw(JPADrawInfo* i_drawInfo) {
+    ZoneScoped;
```
TakaRikka, line 3114: do these need to be behind target defines? not really familiar with how tracy works

PJB3005: If you mean `TARGET_PC`: I figured it would be too much noise and not that big of a deal.

If you mean `TRACY_ENABLE`: no, the header defines them all to be empty if tracy is disabled.

PR #298, opened 2026-04-09
Fix Shop Cursor Positions

MelonSpeedruns:
Probably also fixes other things that rely on `3Dpos -> 2Dpos`.


PJB3005 (commented):
My intuition from reading this signature is that `viewport` should already be the size of the viewport even after our enlarged size, no? Or is this a case where *some* UIs work in native viewport size, and others just got hacked to have their rendering scaled?

TakaRikka:
is this pr still good? i saw some discussion about it causing other issues

MelonSpeedruns:
Nah those were placebo, it's all good!

PR #299, opened 2026-04-09
Fixes map performance even if it's drawn on screen

MelonSpeedruns:
Fixes map performance even if it's drawn on screen.

### DIFF
--- a/src/d/d_map.cpp
+++ b/src/d/d_map.cpp
@@ -542,6 +542,10 @@ void renderingAmap_c::rendering(dDrawPath_c::poly_class const* i_poly) {
 bool renderingAmap_c::isDrawOutSideTrim() {
     bool rt = false;
 
+    #if TARGET_PC
+    return 0;
+    #endif
+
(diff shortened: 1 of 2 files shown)


src/d/d_map.cpp
```diff
@@ -542,6 +542,10 @@ void renderingAmap_c::rendering(dDrawPath_c::poly_class const* i_poly) {
 bool renderingAmap_c::isDrawOutSideTrim() {
     bool rt = false;
 
+    #if TARGET_PC
+    return 0;
+    #endif
```
TakaRikka: what is this actually skipping?

MelonSpeedruns: This skips rendering the black outline outright. The previous fix (including the Hyrule Field Patch) only skipped portions of it, not all, and only when the map was disabled.

TakaRikka: maybe should be behind the hyrule speedhack define? since we might want the outline eventually if we can properly optimize the map

YunataSavior: I agree with Taka here. See if the hyrule speedhack define works here as well.

PR #301, opened 2026-04-10
Implement Pre-Launch UI

CraftyBoss:
Adds an ImGui window that appears before game actually boots. This is used for configuring currently only the game's ISO path, which is needed for android.


TakaRikka:
lgtm as far as i can tell, but @encounter too since he was the one pushing for the pre launch ui

src/dusk/imgui/ImGuiPreLaunchWindow.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+    ImGui::PushFont(ImGuiEngine::fontLarge);
+    ImGuiTextCenter("Options");
+    ImGui::Separator();
+    ImGui::PopFont();
+
+    auto cursorY = ImGui::GetCursorPosY();
+    float endCursorY = windowSize.y - 100;
+
+    float childWidth = windowSize.x - 400;
```
crowell, line 160: what are these magic 100 and 400?

src/dusk/imgui/ImGuiEngine.cpp
```diff
@@ -32,7 +32,28 @@ bool AssetExists(const std::string& path) {
...
+ImFont* ImGuiEngine::fontExtraLarge;
+ImTextureID ImGuiEngine::duskIcon = 0;
+
+inline ImFont* CreateFont(float size, bool hasFontFile,  const std::string& fontPath) {
+    ImGuiIO& io = ImGui::GetIO();
+
+    ImFontConfig fontConfig{};
+    fontConfig.SizePixels = size;
+    snprintf(static_cast<char*>(fontConfig.Name), sizeof(fontConfig.Name),
+             "Noto Mono Regular, %dpx", static_cast<int>(fontConfig.SizePixels));
```
crowell, line 44: can we refactor this "Noto Mono Regular" out to somewhere, because it's used in multiple places now.

src/dusk/imgui/ImGuiPreLaunchWindow.cpp
CraftyBoss: padding for the child window, so its not the same size as the full size window which imo looks a little nicer.

src/dusk/imgui/ImGuiEngine.cpp
CraftyBoss: Was thinking of fully doing this, I ended up making a half refactor with the CreateFont inline func I added.

CraftyBoss:
something I should mention, atm the icon that shows in this UI is loaded from inside the build directory, which means this PR doesn't actually have the asset and will not show in the menu. Unsure how we want to handle ImGui assets so any input on that would be helpful.

src/dusk/imgui/ImGuiPreLaunchWindow.cpp
crowell: perhaps percentage based would be better then?

CraftyBoss: that would probably be far better, ill see what i can do

encounter:
Yes, it looks good. I'm working from this PR as a base to implement the rest of the imgui stuff I want to do

PR #309, opened 2026-04-10
disable fullscreen on macOS to prevent crashes

crowell:


encounter:
Closing in favor of the upstream dawn fix https://dawn-review.googlesource.com/c/dawn/+/302117

PR #342, opened 2026-04-12
Add Move Link tool from decompgz

Pheenoh:


MelonSpeedruns:
Maybe only enable it as part of `mDoMain::developmentMode` ? Since it's behind a hotkey I don't think it needs its own checkbox in the enhancements menu.

src/d/actor/d_a_alink.cpp
```diff
@@ -18048,17 +18049,9 @@ int daAlink_c::execute() {
...
-        if (l_debugMode) {
-            l_debugMode = FALSE;
-        } else {
-            l_debugMode = TRUE;
-        }
-    }
 
-    if (l_debugMode) {
+    if (dusk::getTransientSettings().moveLinkActive && daPy_getPlayerActorClass() == this) {
+        isTrigDebugMoveInput = TRUE;
```
TakaRikka, line 18054: maybe use `#if TARGET_PC` ifdefs to specific newly added pc code separate from the original. applies to the other changes too

src/dusk/imgui/ImGuiConsole.cpp
```diff
@@ -202,6 +203,19 @@ namespace dusk {
...
+
+        if (getSettings().game.enableMoveLink) {
+            static bool comboHeld = false;
+            constexpr u32 combo = PAD_TRIGGER_L | PAD_TRIGGER_R | PAD_BUTTON_Y;
+            u32 rawHold = mDoCPd_c::getHold(PAD_1);
+            bool held = (rawHold & combo) == combo;
+            if (held && !comboHeld) {
+                getTransientSettings().moveLinkActive = !getTransientSettings().moveLinkActive;
+            }
+            comboHeld = held;
```
TakaRikka: is all this logic needed over just checking like `getHoldL + getHoldR + getTrigY`?

include/dusk/settings.h
```diff
@@ -71,6 +71,7 @@ struct UserSettings {
 
         // Controls
         ConfigVar<bool> enableTurboKeybind;
+        ConfigVar<bool> enableMoveLink;
```
TakaRikka: do we want to have vars dedicate to things like this, or just have it be part of enabling development mode?

src/dusk/imgui/ImGuiConsole.cpp
Pheenoh: copy pasta from decompgz, will see if i can reduce it down

PR #343, opened 2026-04-12
Initial implementation of gyro aim

Irastris:
Adds gyro aiming when using a supported controller. Configuration is under a new "Input" submenu (with further reorganizing to come in a subsequent PR) where it can be toggled, sensitivity adjusted, and pitch/yaw inverted. Sensitivity is properly scaled when using Hawkeye, and Mirror Mode is accounted for.

Needs testing with more controllers as I only have a DualSense. The values I chose for the deadband and smoothing feel fantastic on it, but I should probably just expose them in the configuration too, will await others' opinion on that.

### DIFF
--- a/src/dusk/imgui/ImGuiMenuEnhancements.cpp
+++ b/src/dusk/imgui/ImGuiMenuEnhancements.cpp
@@ -111,7 +111,7 @@ namespace dusk {
                 config::ImGuiCheckbox("Gyro Aim", getSettings().game.enableGyroAim);
                 if (ImGui::IsItemHovered()) {
                     ImGui::SetTooltip("Enables the gyroscope on supported controllers while aiming the\n"
-                                      "bow, boomerang, hookshot, copy rod, and iron ball.");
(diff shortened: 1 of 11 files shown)


MelonSpeedruns:
Confirmed working on Switch 2 Pro Controller using https://handheldlegend.github.io/procon2tool/
Though the Yaw was inverted with it by default.

src/dusk/imgui/ImGuiMenuEnhancements.cpp
```diff
@@ -107,6 +107,20 @@ namespace dusk {
                 ImGui::EndMenu();
             }
 
+            if (ImGui::BeginMenu("Input")) {
+                config::ImGuiCheckbox("Gyro Aim", getSettings().game.enableGyroAim);
+                if (ImGui::IsItemHovered()) {
+                    ImGui::SetTooltip("Enables the gyroscope on supported controllers while aiming the\n"
+                                      "bow, boomerang, hookshot, copy rod, and iron ball.");
```
TakaRikka: capitalize the item names, and use the official names like `Clawshot`, `Dominion Rod`, and `Ball and Chain`

PR #357, opened 2026-04-13
speedrun timer

qwertyquerty:


TakaRikka:
i dont really understand how this works atm. i think ideally it should have some auto start / auto end checks, or add some customizable hotkeys for starting / ending the timer

qwertyquerty:
Currently it just has auto start and pauses the livesplit time during non IGT sections, and communicates IGT to livesplit via a frame count of IG frames

TakaRikka:
how does getting the timer to start even work? i couldnt get it to do anything

qwertyquerty:
Need to right click livesplit and start the tcp server under control

Might be worth adding instructions for that somewhere in the IMGUI

TakaRikka:
updates:
- enabling "Speedrun Mode" in the Gameplay settings will now display the on-screen timer
- enabling "Speedrun Mode" resets most cheats and other gameplay modifiers and disables the ability to change them while the mode is active
- the timer shows both RTA and IGT (RTA with loads removed) timers
- timer starts automatically on file load / brightness check depending on what your setting for hide TV setting screen is
- timer ends automatically on finishing ganondorf
- timer can be stopped manually with L+R+A+Z
- timer can be reset with L+R+A+Start

note: the IGT calculation seems bugged atm, it needs some work still. additionally may want to look over what else should be disabled while in speedrun mode, i only did the obvious stuff

PR #359, opened 2026-04-13
Sun's Song

MelonSpeedruns:
Allows Wolf Link to howl to change the time of day.

This is ready to be merged, but team discussion needs to happen to figure out if we want this or the DecompGZ Midna option.


TakaRikka:
would like to revisit the controls but for now this seems good to have

PR #361, opened 2026-04-13
Re-equip Ordon Clothes Feature

MelonSpeedruns:
The Ordon Clothes don't save in the save file!

I don't think it should anyway, because of the nature of Dusk, and being able to toggle Enhancements at runtime, we wouldn't want a specific Enhancement to be stored in the save file forever.


TakaRikka:
i'm a little iffy on this one, im not sure if this is really worth having. maybe its more mod territory? can see what the others think

MelonSpeedruns:
Nah I agree with you, just thought I'd make it. ^^

PR #362, opened 2026-04-13
Revert Classic Bloom Hack

MelonSpeedruns:
Feel free to discuss about this, but I think we should keep the Classic bloom as it is on GameCube without any hacks. Then we can have the HD/Dusk presets set the bloom to Enhanced by default.


MelonSpeedruns:
We decided to go against it.

PR #370, opened 2026-04-14
Android building

CraftyBoss:
Adds new _android_ folder to platforms directory, containing code needed to compile an android apk. Should compile fine, however I've had some occasional hiccups during compilation due to disabling all compression for nod. From when it happens, the linker fails to find non existent funcs for compression, but re-configuring cmake and running a compile again makes the error stop. After which it never happens again. Other than that issue, everything seems to be fine barring any android-exclusive issues that will probably pop up occasionally as more people test it.


libs/JSystem/include/JSystem/JAudio2/JASCalc.h
```diff
@@ -11,9 +11,17 @@
 struct JASCalc {
     static void imixcopy(const s16*, const s16*, s16*, u32);
     static void bcopyfast(const void* src, void* dest, u32 size);
+#if ANDROID
```
TakaRikka: maybe `TARGET_ANDROID` instead? unless it has to be that way

CraftyBoss: ANDROID is what gets supplied by the android cmake target, so I could define TARGET_ANDROID in our own cmakelists if that makes readablity better

TakaRikka: its a little more clear and consistent with the target_pc naming at least

CraftyBoss: got it, will make that adjustment then

PR #392, opened 2026-04-15
Clear statics in trill's destroy call

### LINKED ISSUE #302: [Linux] Trill non functional
Can't refill lantern post monkey lantern stealing sequence

Pheenoh:
Fixes #302 

The lantern refill bit I mentioned was just me being a dumb speedrunner, but there was another issue where trill attempted to scoot toward a stale daMyna_LightActor pointer that was now at 0,0,0, resulting in her scooting in mid air off her post. 

On GC/Wii presumably the REL unload resets them so we don't see the issue there

### DIFF
--- a/src/d/actor/d_a_myna.cpp
+++ b/src/d/actor/d_a_myna.cpp
@@ -363,6 +363,18 @@ int daMyna_c::destroy() {
         mpMorf->stopZelAnime();
     }
 
+#ifdef TARGET_PC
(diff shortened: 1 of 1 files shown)


encounter:
This makes me wonder what other dangling data we have from REL unloads

Pheenoh:
> This makes me wonder what other dangling data we have from REL unloads

Yeah we should definitely dig in here, feels unlikely this was the only one

PR #397, opened 2026-04-15
Fix Unloading Gates

SuperDude88:
Fixes CitS/Lakebed gate unloading (only present in TPHD because it doesn't have RELs), bug comment/brief actor description should be updated in decomp


YunataSavior:
Your comment in the cpp code should indicate that this is a TPHD and static-loading bug only (in other words, it doesn't happen on GCN/Wii)

Antidote:
> Your comment in the cpp code should indicate that this is a TPHD and static-loading bug only (in other words, it doesn't happen on GCN/Wii)

Yes, that's true, but dusk *also* static links them, which is why TPHD is relevant.

PR #413, opened 2026-04-17
add uninitialized variable fix

roeming:
Initializes an uninitialized local variable. This variable is added as an offset for a translation matrix, so setting it to zero seems like a reasonable value for that purpose.


src/d/actor/d_a_npc_inko.cpp
```diff
@@ -110,6 +110,10 @@ static int daNpc_Inko_Execute(npc_inko_class* i_this) {
     }
 
     f32 var_f31;
+    
+    #if TARGET_PC
```
PJB3005:
```suggestion
    #if AVOID_UB
```

PR #421, opened 2026-04-18
Custom Meter Colors

Captainkittyca2:
Navigate to Debug -> Meter Colors.

You can alter the color of Lantern Meter, Oxygen Meter, and low Oxygen meter.


src/dusk/imgui/ImGuiBloomWindow.cpp
```diff
@@ -21,6 +23,15 @@ struct BloomOverride {
 
 BloomOverride s_bloomOverride;
 
+void ResetToDefaultMeterColors() {
```
TakaRikka: this doesn't make sense in ImGuiBloomWindow. should be made a member function of ImGuiMenuEnhancements

src/dusk/imgui/ImGuiBloomWindow.cpp
```diff
@@ -123,4 +134,33 @@ void DrawBloomWindow(bool& open) {
...
+    DrawColorEdit("Oxygen Bottom (Low)", s_meterColorsOverride.oxygen2CustomBottom);
+    if (ImGui::MenuItem("Reset meter colors to default")) {
+        ResetToDefaultMeterColors();
+    }
+    ImGui::End();
+}
+
+void ImGuiMenuTools::ShowMeterColorsWindow() {
+    DrawMeterColorsWindow(m_showMeterColors);
+}
```
TakaRikka: all of this too can be simplified as part of ImGuiMenuEnhancements. also should be an enhancement. doesn't make sense as a tool

src/dusk/imgui/ImGuiBloomWindow.hpp
```diff
@@ -1,6 +1,17 @@
...
+    GXColor lanternCustomTop = {230, 170, 0, 255};
+    GXColor lanternCustomBottom = {255, 255, 140, 255};
+    GXColor oxygen1CustomBottom = {200, 200, 255, 255};
+    GXColor oxygen1CustomTop = {80, 180, 255, 255};
+    GXColor oxygen2CustomBottom = {255, 100, 100, 255};
+    GXColor oxygen2CustomTop = {255, 10, 10, 255};
+};
+
+extern MeterColorsOverride s_meterColorsOverride;
+void DrawMeterColorsWindow(bool& open);
```
TakaRikka: move to ImGuiMenuEnhancements. also these seem to all be transient settings, so it won't save your settings for next session? it'd probably make more sense to have these be CVars

src/d/d_meter2_draw.cpp
```diff
@@ -22,6 +22,9 @@
 #include "d/d_pane_class.h"
 #include "dusk/frame_interpolation.h"
 #include <cstring>
+#if Target_PC
```
TakaRikka: should be `TARGET_PC`

src/dusk/imgui/ImGuiBloomWindow.cpp
Captainkittyca2: Ah I thought it fit in debug tab since bloom color modifications were also in debug. I'll change it to Enchanements, thx. Tbh, one of the reasons why I had it in ImGuiBloomWindow is because DrawColorEdit is in there but I'll change DrawColorEdit to be in enhancement area

src/d/d_meter2_draw.cpp
Captainkittyca2: Fixed, thanks.

src/dusk/imgui/ImGuiBloomWindow.cpp
Captainkittyca2: I moved the resetting function to ImGuiMenuEnhancements.cpp

src/dusk/imgui/ImGuiBloomWindow.cpp
Captainkittyca2: I changed it to be in enhancement now.

src/dusk/imgui/ImGuiBloomWindow.hpp
Captainkittyca2: I tried for hours to create a ConfigVar for GXColor and failed and I also tried making multiple u8 ConfigVars but ran into diffiucilty with using them with EditColor4 function (or DrawColorEdit function) so unfortunately, still can't save the values, but otherwise, everything is good to go.

PR #425, opened 2026-04-18
Interpolation fixes for d_meter_haihai, d_msg_out_font and d_meter_button

### LINKED ISSUE #419: Option Screen locked to FPS speed
https://github.com/user-attachments/assets/4e7f48f0-728d-4117-981c-259407ab75c3

Pheenoh:
Closes #419

### DIFF
--- a/src/d/d_meter_button.cpp
+++ b/src/d/d_meter_button.cpp
@@ -16,6 +16,7 @@
 #include "d/d_msg_out_font.h"
 #include "d/d_msg_string.h"
 #include "d/d_pane_class.h"
+#include "dusk/frame_interpolation.h"
 #include <cstring>
 
 #if VERSION == VERSION_GCN_JPN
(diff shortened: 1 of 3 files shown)


Pheenoh:
Also closes #366

MelonSpeedruns:
LGTM!

Irastris:
I want the header includes ifdef'd first, the only reason they may not be in other TUs is because that was work done prior to the interp branch merging with main and it wasn't cleaned up appropriately.

Pheenoh:
> I want the header includes ifdef'd first, the only reason they may not be in other TUs is because that was work done prior to the interp branch merging with main and it wasn't cleaned up appropriately.

Can you say more why you want the headers guarded?

Irastris:
To save ourselves extra work later. @TakaRikka has stated their dream for Dusk is that it could still be compiled for GameCube later on. If they no longer have this goal in mind then I don't care, but until then we shouldn't intentionally ignore this stuff.

Pheenoh:
> To save ourselves extra work later. @TakaRikka has stated the dream for Dusk is that it could still be compiled for GameCube later on. If they no longer have this mind in goal then I don't care, but until then, we shouldn't intentionally ignore this stuff.

How would guarding the headers reduce work exactly? The call sites are guarded already and the header only has declarations. I'm not against adding this it's just guarding headers with declarations only is a weird pattern to me

Irastris:
I haven't given it that much thought, I'm just following the pattern of other TUs. If you don't view it as necessary then fine.

TakaRikka:
yeah i mean maintaining gc compatibility is mostly a pipedream. it'd be cool, but ultimately the only thing that really matters here is the pc side of things

PR #427, opened 2026-04-18
Fix file select fade not rendering with frame interpolation

Pheenoh:
mpFadePict on dFile_select_c is drawn directly rather than through the display list system. Since interpolation defers cAPIGph_Painter's execution to after process draws, the direct draw gets wiped when the painter clears the framebuffer, and the fade transition between file naming screens never appears. Adding a new dDlst_FileSelFade_c wrapper that inherits from dDlst_base_c so the fade is drawn through the display list and survives the deferred painter

### DIFF
--- a/include/d/d_file_select.h
+++ b/include/d/d_file_select.h
@@ -10,6 +10,7 @@
 #include "JSystem/J3DGraphLoader/J3DAnmLoader.h"
 
 class dFile_info_c;
+class J2DPicture;
 
 class dDlst_FileSel_c : public dDlst_base_c {
 public:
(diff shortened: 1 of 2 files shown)


TakaRikka:
i dont mind this, but is this a deeper issue with interpolation that needs to be fixed rather than worked around?

MelonSpeedruns:
> i dont mind this, but is this a deeper issue with interpolation that needs to be fixed rather than worked around?

Fully agree with that.

Pheenoh:
> i dont mind this, but is this a deeper issue with interpolation that needs to be fixed rather than worked around?

I think it's fine for now because I couldn't find any other direct draw cases like this. But I think it's worth keeping this in the back of our heads as we find more interp issues, we might need to rethink where and how cAPIGph_Painter get executed if we come across more of these

TakaRikka:
sounds good to me

Irastris:
The possibility worries me, especially as getting water refraction to work continues to evade me, but the deferred painter has proved the most stable solution so far. I fear that any other solution would likely require significant fixups of the entire game on a level I doubt I'd have the technical skill to achieve.

Pheenoh:
> The possibility worries me, especially as getting water refraction to work continues to evade me, but the deferred painter has proved the most stable solution so far. I fear that any other solution would likely require significant fixups of the entire game on a level I doubt I'd have the technical skill to achieve.

Yeah I'm more inclined to continue to try and get the deferred painter to work unless it ends up becoming a huge hassle

PR #453, opened 2026-04-20
Mod loader

qwertyquerty:


include/dusk/mod_api.h
PJB3005, line 1: Given that it seems implausible we'll ever have non-C++ mods, does this really need to be a C header? It would probably be a lot more convenient with proper C++ types and such.

cmake/dusk_imgui_ctx.cpp
```diff
@@ -0,0 +1,5 @@
+#include "imgui.h"
+
+extern "C" void dusk_mod_set_imgui_ctx(void* ctx) {
```
PJB3005: Uh, wouldn't it be way better if the mod just linked against Dear Imgui in Dusk?

include/dusk/mod_api.h
```diff
@@ -0,0 +1,58 @@
...
+#else
+#  define DUSK_MOD_EXPORT __attribute__((visibility("default")))
+#endif
+
+// Place this once at file scope in your mod to declare the minimum API version required.
+// The loader will refuse to initialize the mod if the engine's API version is older.
+#define DUSK_REQUIRE_API_VERSION \
+    DUSK_MOD_EXPORT uint32_t mod_api_version = DUSK_MOD_API_VERSION;
+
+typedef struct DuskModAPI {
```
PJB3005: Probably best to explicitly list this as "mod API v1" and have games refer to it via an alias. That way we don't need an awkward shuffle on the first ABI break.

src/dusk/imgui/ImGuiMenuTools.cpp
PJB3005, line 1: You probably did not mean to commit this in this branch?

src/dusk/gx_helper.cpp
```diff
@@ -0,0 +1,13 @@
+#include "dusk/gx_helper.h"
+
+#ifdef TARGET_PC
+GXTexObjRAII::~GXTexObjRAII() { GXDestroyTexObj(this); }
+void GXTexObjRAII::reset() { GXDestroyTexObj(this); }
+#endif
```
PJB3005: These ifdefs are unecessary

src/dusk/hook_system.cpp
```diff
@@ -0,0 +1,155 @@
...
+    return s;
+}
+static std::unordered_map<uintptr_t, void*>& installed() {
+    static std::unordered_map<uintptr_t, void*> s;
+    return s;
+}
+
+// Follow E9/FF25 chains to skip MSVC incremental-link and import stubs
+static void* resolveImportThunk(void* addr) {
+#if _WIN32
```
PJB3005: This needs to check if it's x64 specifically, as I doubt this works on ARM64.

src/dusk/hook_system.cpp
```diff
@@ -0,0 +1,155 @@
...
+};
+
+static std::unordered_map<uintptr_t, HookSlot>& registry() {
+    static std::unordered_map<uintptr_t, HookSlot> s;
+    return s;
+}
+static std::unordered_map<uintptr_t, void*>& installed() {
+    static std::unordered_map<uintptr_t, void*> s;
+    return s;
+}
```
PJB3005: Do these actually need to be delay initialized like this?

src/dusk/modding/mod_loader.cpp
```diff
@@ -0,0 +1,412 @@
...
+    while (!s.empty() && (s.back() == '\r' || s.back() == '\n'))
+        s.pop_back();
+    return s;
+}
+static constexpr const char* k_libExt = ".dll";
+
+#else
+#include <dlfcn.h>
+static void* pl_dlopen(const std::filesystem::path& p) {
+    return dlopen(p.c_str(), RTLD_LAZY | RTLD_LOCAL);
```
PJB3005: Best to add `RTLD_DEEPBIND` here to avoid issues.

src/dusk/mod_loader.cpp
```diff
@@ -0,0 +1,412 @@
...
+    return e ? e : "(unknown error)";
+}
+#if defined(__APPLE__)
+static constexpr const char* k_libExt = ".dylib";
+#else
+static constexpr const char* k_libExt = ".so";
+#endif
+#endif
+
+static dusk::LoadedMod* g_currentMod = nullptr;
```
PJB3005: This isn't thread-safe as is and I'd really just prefer if the "current mod" was explicitly an argument passed into the callbacks, at least at this level of the code.

src/dusk/mod_loader.cpp
```diff
@@ -0,0 +1,412 @@
...
+    std::string s(vsnprintf(nullptr, 0, fmt, ap2), '\0'); va_end(ap2);
+    vsnprintf(s.data(), s.size() + 1, fmt, ap); va_end(ap);
+    DuskLog.error("[{}] {}", modName(), s);
+}
+
+static void* cb_load_resource(const char* relative_path, size_t* out_size) {
+    if (out_size)
+        *out_size = 0;
+    if (!g_currentMod || !relative_path)
+        return nullptr;
```
PJB3005: This should probably at least log an error?

src/dusk/mod_loader.cpp
```diff
@@ -0,0 +1,412 @@
...
+
+    mz_zip_archive zip{};
+    if (!mz_zip_reader_init_file(&zip, g_currentMod->mod_path.c_str(), 0)) {
+        DuskLog.warn("[{}] load_resource: could not open {}", g_currentMod->name,
```
PJB3005: nit: should be error log level

src/dusk/mod_loader.cpp
```diff
@@ -0,0 +1,412 @@
...
+                     g_currentMod->mod_path);
+        return nullptr;
+    }
+    std::string entry = std::string("res/") + relative_path;
+    size_t sz = 0;
+    void* data = mz_zip_reader_extract_file_to_heap(&zip, entry.c_str(), &sz, 0);
+    mz_zip_reader_end(&zip);
+    if (!data) {
+        DuskLog.warn("[{}] load_resource: '{}' not found in zip", g_currentMod->name, entry);
```
PJB3005: ditto

src/dusk/mod_loader.cpp
```diff
@@ -0,0 +1,412 @@
...
+}
+
+static void cb_register_tab_content(void (*draw_fn)(void*), void* userdata) {
+    if (g_currentMod && draw_fn)
+        g_currentMod->tab_content.push_back({draw_fn, userdata});
+}
+
+static void cb_service_publish(const char* name, void* ptr) {
+    if (name) {
+        g_services[name] = ptr;
```
PJB3005: This should probably log an error if two separate mods publish the same service?

src/dusk/mod_loader.cpp
```diff
@@ -0,0 +1,412 @@
...
+        if (g_currentMod) {
+            g_currentMod->load_failed = true;
+        }
+    }
+}
+
+namespace dusk {
+
+ModLoader& ModLoader::instance() {
+    static ModLoader inst;
```
PJB3005: Again, does this need a lazy init?

src/dusk/mod_loader.cpp
```diff
@@ -0,0 +1,412 @@
...
+}
+
+static void* cb_load_resource(const char* relative_path, size_t* out_size) {
+    if (out_size)
+        *out_size = 0;
+    if (!g_currentMod || !relative_path)
+        return nullptr;
+
+    mz_zip_archive zip{};
+    if (!mz_zip_reader_init_file(&zip, g_currentMod->mod_path.c_str(), 0)) {
```
PJB3005: Is it not very inefficient to re-open the zip file every single time?

src/dusk/mod_loader.cpp
```diff
@@ -0,0 +1,412 @@
...
+        mz_zip_archive_file_stat stat{};
+        if (!mz_zip_reader_file_stat(&zip, i, &stat))
+            continue;
+        if (mz_zip_reader_is_file_a_directory(&zip, i))
+            continue;
+        if (fs::path(stat.m_filename).extension() == k_libExt) {
+            dllEntry = stat.m_filename;
+            break;
+        }
+    }
```
PJB3005: This system needs to account for multiple DLLs for separate architectures.

src/dusk/mod_loader.cpp
```diff
@@ -0,0 +1,412 @@
...
+                     modPath.filename().string());
+        return;
+    }
+
+    const fs::path cacheDir = m_modsDir / ".cache" / modPath.stem();
+    std::error_code ec;
+    fs::create_directories(cacheDir, ec);
+
+    const fs::path dllCachePath = cacheDir / fs::path(dllEntry).filename();
+    if (!mz_zip_reader_extract_file_to_file(&zip, dllEntry.c_str(), dllCachePath.string().c_str(),
```
PJB3005: `.string()` is probably not gonna handle Unicode properly.

src/dusk/mod_loader.cpp
```diff
@@ -0,0 +1,412 @@
...
+    mod.api.service_get     = cb_service_get;
+}
+
+void ModLoader::tryLoadDusk(const std::filesystem::path& modPath) {
+    namespace fs = std::filesystem;
+
+    std::string metaName, metaVersion, metaAuthor, metaDescription;
+    {
+        mz_zip_archive zip{};
+        if (mz_zip_reader_init_file(&zip, modPath.string().c_str(), 0)) {
```
PJB3005: `.string()` is probably not gonna handle Unicode properly.

src/dusk/mod_loader.cpp
```diff
@@ -0,0 +1,412 @@
...
+                    mz_free(jsonData);
+                    DuskLog.warn("ModLoader: bad mod.json in {}: {}", modPath.filename().string(),
+                                 e.what());
+                }
+            }
+        }
+    }
+
+    mz_zip_archive zip{};
+    if (!mz_zip_reader_init_file(&zip, modPath.string().c_str(), 0)) {
```
PJB3005: `.string()` is probably not gonna handle Unicode properly.

src/dusk/modding/mod_loader.cpp
```diff
@@ -0,0 +1,412 @@
...
+
+    void* handle = pl_dlopen(dllCachePath);
+    if (!handle) {
+        DuskLog.error("ModLoader: failed to open {}: {}", dllCachePath.string(), pl_dlerror());
+        return;
+    }
+
+    LoadedMod mod;
+    mod.mod_path = fs::absolute(modPath).string();
+    mod.dir = fs::absolute(cacheDir).string();
```
PJB3005: `.string()` doesn't handle Unicode properly.

src/dusk/mod_loader.cpp
PJB3005: Do we need to rename the mod file to ensure two mods with the same dll name don't cause conflicts? Or is that overkill.

PJB3005 (changes requested):
Didn't review the hooking code or any of the build stuff yet

src/dusk/mod_loader.cpp
qwertyquerty: why not

src/dusk/mod_loader.cpp
qwertyquerty: `const fs::path cacheDir = fs::path("mods") / ".cache" / modPath.stem();`

src/dusk/mod_loader.cpp
PJB3005: > why not

I don't think it'll convert to UTF-8 on Windows.

src/dusk/imgui/ImGuiMenuTools.cpp
qwertyquerty: ya but i cherry picked it to main so its fine

src/dusk/mod_loader.cpp
PJB3005: The actual final filename needs to be different too. I don't think Windows will be happy having two DLLs with the same filename but different paths loaded.

qwertyquerty: ah ok

qwertyquerty: still might be overkill, the dll should have the name of the .dusk and you wouldn't be able to put two identically named .dusk in the same folder unless you specifically renamed them and then you the user are breaking things by choice

src/dusk/mod_loader.cpp
qwertyquerty: good point

SuperDude88:
Superseded by #2185

PR #458, opened 2026-04-20
Autosave Feature

MelonSpeedruns:
Needs to be tested in depth to make sure it doesn't corrupt save files at any point!


TakaRikka:
are we thinking this is safe now or is it still too experimental?

MelonSpeedruns:
@PJB3005 did a run with it and it was fine. It just autosaves when loading a save state.

src/d/d_menu_save.cpp
```diff
@@ -1335,9 +1335,6 @@ void dMenu_save_c::memCardDataLoadWait() {
 }
 
 void dMenu_save_c::dataWrite() {
-    int stageNo = dStage_stagInfo_GetSaveTbl(dComIfGp_getStageStagInfo());
-
-    dComIfGs_putSave(stageNo);
```
TakaRikka, line 1340: whats the reason for this removal?

MelonSpeedruns: I removed it because it seemed to change the stage you would respawn into to the wrong one, but that was when I was saving mid transition. I will check if the vanilla game does that line today and re-add it if it does.

MelonSpeedruns: Yeah this was a remain of me trying saving stuff, so I restored it, good catch!

TakaRikka:
lgtm but does this need to be updated now that rmlui is merged?

MelonSpeedruns:
it will need to be updated when rmlui gets Toasts, but right now it doesn't support them AFAIK.

encounter:
needs to be ported to the new gameplay menu, that exists

PR #485, opened 2026-04-22
Rework interpolation pacing; interpolate every frame

encounter:
This improves frame pacing when interpolating by interpolating every presentation frame based on when we start to draw. One way to think about it: We use simulation ticks to add the next 33.3ms to the "timeline", and on draw, "sample" (via interpolation) the current time on the timeline. This yields seemingly-perfect pacing at any framerate.

One caveat: input is still tied to _when_ we run the simulation ticks, which could be up to 33ms delayed if we end up sampling close to the end of the current timeline. We may be able to improve this another way.

Before: https://discord.com/channels/1446645861529550869/1446648797701738518/1496364951768399982
After: https://discord.com/channels/1446645861529550869/1446648797701738518/1496364179005771818

### DIFF
--- a/include/dusk/frame_interpolation.h
+++ b/include/dusk/frame_interpolation.h
@@ -16,6 +16,7 @@ void ensure_initialized();
 
 void begin_record();
 void end_record();
+void begin_sim_tick();
 void begin_frame(bool enabled, bool is_sim_frame, float step);
 void interpolate();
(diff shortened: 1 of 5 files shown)


Antidote:
Your caveat should only really become a noticeable issue if it starts to dip below real 30FPS. Simulated frames should, in theory, feel correct otherwise.

PR #488, opened 2026-04-22
Items Don't Despawn

Captainkittyca2:
Makes dropped items NOT disappear when timer expires. They stay and can be collected still.

Gameplay -> Indefinite Item drops

### DIFF
--- a/include/dusk/settings.h
+++ b/include/dusk/settings.h
@@ -47,6 +47,7 @@ struct UserSettings {
 
     struct {
         // QoL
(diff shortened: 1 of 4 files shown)


MelonSpeedruns:
Looks good to me, maybe just rename it to "Drops Never Despawn" as the word "Indefinite" might be hard to understand for some users.

Captainkittyca2:
Done! Renamed to Items Don't Despawn as discussed in Discord server and also now under Cheats tab. And changed the description to Tom's description from the discord server lol

PR #492, opened 2026-04-22
Gamepad Colors for DS4/DS5

MelonSpeedruns:
Changes colors with the following:
- If Midna wants your attention
- Your current form
- Your current tunic
- If you are in a twilight zone or not.


TakaRikka:
maybe this stuff should be put in its own file rather than just stuffed in f_ap_game?

MelonSpeedruns:
> maybe this stuff should be put in its own file rather than just stuffed in f_ap_game?

yeah, good idea!

PR #514, opened 2026-04-24
Disc verification

dooplecks:


encounter:
Looks good so far. I'm thinking we should persist either a map of paths to known valid states, or just store "discValidated" in config and reset it when we change ISO paths. This way the user doesn't have to validate every time they run Dusk.

Also, may be useful to have a "Skip Disc Validation" in the Interface settings

dooplecks:
Made some changes to the UX to make it a bit more seamless:
- It will always attempt verification after selecting a file.
  - If it passes, the progress modal simply closes.
  - Upon error, a warning modal will be shown to indicate the type of error.
  - The user may cancel the verification, but this will also trigger the warning modal.
- The settings page indicates the results of the last verification attempt (pass/fail/unknown).

TakaRikka:
would this be better to wait until the rmlui implementation is finished?

PR #517, opened 2026-04-24
Clamp max LOD for Wolf Link and Midna eyes

encounter:


encounter:
Does not include shadow Midna yet

PR #532, opened 2026-04-24
Even faster tears to match TPHD

MelonSpeedruns:


TakaRikka:
does this actually match tphd? i got the values i did there from looking at the function in ghidra

Pheenoh:
> does this actually match tphd? i got the values i did there from looking at the function in ghidra

Haven't checked if this matches but our current impl was definitely still slower than TPHD

MelonSpeedruns:
Huh? Well in TPHD, you can stand on the bug, and as soon as it explodes, you collect it. Setting the timer to 0 matches that, but might not be the same that TPHD does?

TakaRikka:
weird, maybe theres some other stuff i missed from their original impl. reversing tphd sucks ass though, so if this works then sure

PR #559, opened 2026-04-26
phong lighting model

qwertyquerty:


jlrouzies-fr:
could you post some comparison pictures?

also I wonder, could effects like RTGI be achieved?

PR #565, opened 2026-04-26
cLib_offsetPos param aliasing fix

### LINKED ISSUE #537: Ghost riders in horseback go the wrong way

Pheenoh:
MWCC schedules vec members into registers before any stores to pdest. On PC pdest gets written to first which is an issue when pdest and vec are the same pointer (only happens in a few places). Storing them in temp variables first fixes this

Closes #537 

Also fixes an issue in the Argorok cutscene where the camera was in the wrong place when focusing on the jewel (happens at the start of phase 2)

### DIFF
--- a/src/SSystem/SComponent/c_lib.cpp
+++ b/src/SSystem/SComponent/c_lib.cpp
@@ -468,9 +468,20 @@ s16 cLib_targetAngleX(cXyz const* lhs, cXyz const* rhs) {
 void cLib_offsetPos(cXyz* pdest, cXyz const* psrc, s16 angle, cXyz const* vec) {
     f32 cos = cM_scos(angle);
     f32 sin = cM_ssin(angle);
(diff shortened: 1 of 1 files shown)


src/SSystem/SComponent/c_lib.cpp
```diff
@@ -468,9 +468,18 @@ s16 cLib_targetAngleX(cXyz const* lhs, cXyz const* rhs) {
 void cLib_offsetPos(cXyz* pdest, cXyz const* psrc, s16 angle, cXyz const* vec) {
     f32 cos = cM_scos(angle);
     f32 sin = cM_ssin(angle);
+#if TARGET_PC
```
PJB3005: I'd definitely put a comment here explaining that this is a compiler bug. If you want extra credit I guess change the condition to be `!__MWERKS__`

PJB3005 (approved):

PR #570, opened 2026-04-27
fix main map for mirror mode

TakaRikka:
this isn't fully done yet, the map is still off center and selection regions are still inverted


TakaRikka:
all main issues with it fixed now

PR #583, opened 2026-04-28
Email me and Heavy Hitter Achievements

TheLastPocket:
Achievement for reading a letter during beast ganon


TheLastPocket:
Added the heavy hitter achievement for wearing the iron boots in the credits

PR #590, opened 2026-04-29
Improve map outline rendering

dooplecks:
Before and after:

map-outline-comparison

### DIFF
--- a/src/d/d_map_path.cpp
+++ b/src/d/d_map_path.cpp
@@ -44,33 +44,34 @@ void dMpath_n::dTexObjAggregate_c::create() {
         const auto center = kMapCircleSize / 2.0f;
         const auto radiusSq = center * center;
         const auto blocksAcross = kMapCircleSize >> 3;
+        const auto totalPixels = sizeof(hqCircleData);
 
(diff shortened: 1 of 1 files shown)


src/d/d_map_path.cpp
```diff
@@ -32,6 +33,48 @@ void dMpath_n::dTexObjAggregate_c::create() {
         JUT_ASSERT(74, image->magFilter == GX_NEAR);
         mDoLib_setResTimgObj(image, mp_texObj[lp1], 0, NULL);
     }
+
+#if TARGET_PC
+    auto hqCircle = JKR_NEW TGXTexObj();
```
encounter, line 38: Should definitely keep the texobj around (it can be static)

dooplecks: I tried this but `dMpath_n::dTexObjAggregate_c::remove` attempts to free it and crashes. This seemed like the most straightforward option.

encounter: I misread, I thought this was in draw. All good

encounter (approved):

PR #599, opened 2026-04-30
More Achievements

qwertyquerty:
Going for a decent set of 50 by launch


src/dusk/achievements.cpp
```diff
@@ -326,6 +666,87 @@ std::vector<AchievementSystem::Entry> AchievementSystem::makeEntries() {
             },
             {}
         },
+        {
+            {
+                "email_me",
```
TakaRikka, line 716: i get what the joke is but it seems a little obtuse for an achievement

src/dusk/achievements.cpp
```diff
@@ -411,34 +832,85 @@ std::vector<AchievementSystem::Entry> AchievementSystem::makeEntries() {
...
             },
             {}
         },
         {
             {
-                "heavy-hitter",
-                "Heavy Hitter",
-                "Wear the Iron Boots during the end credits.",
-                AchievementCategory::Misc,
+                "null_item",
```
TakaRikka, line 903: kinda mixed on this one, maybe ask some other runners on their thoughts? and on the achievement list in general

src/dusk/achievements.cpp
```diff
@@ -411,34 +832,85 @@ std::vector<AchievementSystem::Entry> AchievementSystem::makeEntries() {
...
+                        a.progress = 1;
+                        break;
+                    }
+                }
+            },
+            {}
+        },
+        {
+            {
+                "stallord_skip",
```
TakaRikka, line 924: idk about this one, i'd kinda like more interesting glitch achievements rather than just "do this hard trick". BiT and EMS i can accept cause they're so ubiquitous, but this one feels meh to me. i'm thinking more along the lines of "beat spr before clearing lanayru twilight" or "save hidden village before going to goron mines" or stuff like that

TakaRikka (commented):
some other thoughts on previously added achievements:
- for the 3 heart CoO one, maybe it should be relative to your max health? like do it without losing more than 3 hearts. just a thought
- for earliest master sword, requiring a rupee slide for an achievement might be a bit much. idk, maybe another one to get other runners thoughts on

src/dusk/achievements.cpp
qwertyquerty: I agree that sequence break ones are good ideas, I'll think about it more

src/dusk/achievements.cpp
qwertyquerty: The idea was to make a couple of them gimmicky / easy

TheLastPocket:
Made a PR that adds 2 new achievements to this branch.
 https://github.com/TwilitRealm/dusk/pull/644

PR #612, opened 2026-05-01
Increase gameHeap size by an order of magnitude

PJB3005:
Why not?


TakaRikka:
@encounter you have any objections to this?

encounter:
I would really like to not forget we did this (a big scary comment would be nice) but other than that sure

PR #616, opened 2026-05-01
Emulate Surround option

qwertyquerty:


ThreeDeeJay:
I'm not sure if I'm missing something, but if I enable spatial audio and surround, I still don't hear HRTF
image
image
It does change the audio a bit (quieter) but I don't think surround virtualization is working properly.

By the way, does it decode the matrix surround (Dolby Pro-Logic) signal?
[Dolphin uses FreeSurround](https://airtable.com/appayGNkn3nSuXkaz/shrZP6E5xqQjsvplj?eCzZW=recmf3WVq9MPC5La9) for decoding and it's somewhat decent with an external headphone virtualizer like HeSuVi:

https://github.com/user-attachments/assets/1272d3a3-f04a-4f0d-acd3-5e74a7d30e13

Though I think it would be better if it completely bypassed matrix surround encoding/decoding which degrades quality and adds delay and artifacts. Or better yet, perhaps it could use [OpenAL Soft](https://github.com/kcat/openal-soft) for full 3D object-based mixing for proper 3D spatial audio (or surround, Atmos, etc.). 👀👌

PR #637, opened 2026-05-03
Prevent trim and fader gaps around window edge

### LINKED ISSUE #490: Fader and letterboxing masks aren't flush with top of viewport in some situations
Probable reproduction requirements:
- 1x internal resolution
- force 4:3 aspect ratio
- GNOME with PaperWM (lol?)
  - it overrides the window size/position immediately on launch, which might be the real trigger
  - not sure if it happens with other (scrolling-)tiling window managers

Image

Can be fixed by simply resizing the window or changing IR.

### LINKED ISSUE #635: Black bars are not fully going to the sides of the screen with Force 4:3 aspect ratio.
Image

dooplecks:
Fixes #635 and fixes #490.

Not sure if the `J2DOrthoGraph` change is too aggressive.

### DIFF
--- a/libs/JSystem/src/J2DGraph/J2DGrafContext.cpp
+++ b/libs/JSystem/src/J2DGraph/J2DGrafContext.cpp
@@ -25,7 +25,13 @@ void J2DGrafContext::setPort() {
     if (bounds.i.y < 0.0f) {
         bounds.i.y = 0.0f;
     }
+#ifdef TARGET_PC
+    // HACK: compensate for viewport being slightly smaller than scissor region in some cases
+    GXSetViewport(bounds.i.x - 0.5f, bounds.i.y - 0.5f, bounds.getWidth() + 1.0f,
+        bounds.getHeight() + 1.0f, 0.0f, 1.0f);
+#else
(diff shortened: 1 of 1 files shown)


TakaRikka:
i am a little worried this might be an aggressive change, maybe its fine? would like someone else to also look at this

MelonSpeedruns:
I think it's fine, yeah it's aggressive, but it's behind the TARGET_PC define.
Maybe a comment to mention it might need updating in the future?

dooplecks:
I've simplified the fix into a one-line change in `J2DGrafContext`. Instead of adjusting orthographic projection bounds, we make the viewport one pixel wider and taller. Still hacky and ripe for side effects.

The core of the issue seems to lie with how Aurora scales logical viewports and scissor regions. For example, when forcing 4:3 with window dimensions of 1493x832, Aurora will map `GXSetScissor(0, 0, 608, 448)` to an actual scissor region of `(181, 0, 1131, 832)`, and it will map `GXSetViewport(0, 0, 608, 448, ...)` to roughly `(181.929, 0, 1129.14, -832, ...)`. Note that the viewport is smaller than the scissor region, so vertices with normalized coordinates of -1/+1 don't actually meet the edge of the scissor region when scaled to screen coordinates.

dooplecks:
Pending encounter/aurora#155 for a more proper fix.

PR #638, opened 2026-05-03
Display the real title screen behind the prelaunch menu

Irastris:
This makes the game start execution as soon as the disc image is valid (or at startup if a valid path was already set), displaying a halted title screen behind prelaunch until the user selects "Play".

**Pros**:

- It looks so awesome

**Cons**:

- This makes both disc image and save type swaps require a restart. How often do we anticipate people doing that for it to be that much of an inconvenience, though?


src/dusk/ui/modal.cpp
```diff
@@ -0,0 +1,75 @@
...
+                callback();
+            }
+        });
+        mButtons.push_back(std::move(btn));
+    }
+}
+
+bool Modal::focus() {
+    if (!mButtons.empty()) {
+        return mButtons.back()->focus();
```
encounter: This should probably be front(), and presets window should override it to back() (it was a hack to select the Dusk preset by default)

encounter:
I'd say we merge it after fixing the conflict and testing the restart code on Windows

PR #640, opened 2026-05-03
Fix OSReport crash on linux, harden FormatToString against malformed strings

Antidote:


TakaRikka:
seems like this is broken on msvc

PR #642, opened 2026-05-03
Add fly mode option to debug cam

Pheenoh:
Add the debug fly cam from gz. also removing free look per craftyboss


src/d/d_camera.cpp
```diff
@@ -7474,7 +7479,99 @@ bool dCamera_c::test2Camera(s32 param_0) {
...
+        }
+        return false;
+    }
+
+    dEvt_control_c* event = dComIfGp_getEvent();
+    if (event == nullptr) {
+        return false;
+    }
+
+    event->mEventStatus = 1;
```
TakaRikka: just curious, did you test if this messes up other events like it used to gz? like when you'd try to use move link after calling midna and it'd bring her up cause the event state was still stored.

Pheenoh: Yeah it was busted here too. I just disabled the feature if the player is in an event or the pause flag is set

PR #644, opened 2026-05-03
Spin Achievements

TheLastPocket:
Addendum to Qwerty's 50 Achivements

### DIFF
--- a/src/d/actor/d_a_alink_cut.inc
+++ b/src/d/actor/d_a_alink_cut.inc
@@ -9,6 +9,10 @@
 
 #include "dusk/settings.h"
 
+#ifdef TARGET_PC
+#include "dusk/achievements.h"
+#endif
+
 enum daAlink_CutNmParamType {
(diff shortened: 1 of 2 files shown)


TakaRikka:
personally kinda meh on these achievements, but i'll let some others chime in their thoughts

tomlube:
> personally kinda meh on these achievements, but i'll let some others chime in their thoughts

agreed

TheLastPocket:
> personally kinda meh on these achievements, but i'll let some others chime in their thoughts

I really like them as the player can earn these at any time, needs concentration/focus, and needs to learn to get good at the quick spin, which _is_ a game mechanic

TheLastPocket:
I might add that it's impossible to complete these without back-to-back quick spins, which are pretty hard to do

TheLastPocket:
Ok what if we got rid of the 5 spins one and kept the 10 spins one. Or if you feel like its too easy up it to 15

PR #660, opened 2026-05-04
Fix a UB in Morpheel

MelonSpeedruns:


TakaRikka:
lgtm but what does this actually fix out of curiosity?

src/d/actor/d_a_e_oct_bg.cpp
PJB3005, line 519: Wouldn't it be better if you put the added line here in an else? So it's more clear where it applies.

PJB3005:
> lgtm but what does this actually fix out of curiosity?

If nothing else, at least a runtime crash when compilers are configured with UB detection.

MelonSpeedruns:
> lgtm but what does this actually fix out of curiosity?

Doesn't fix anything in builds, but I had this block the game when fighting morpha in Debug mode.

PJB3005 (approved):

PR #684, opened 2026-05-05
Remove post-trim scissor

### LINKED ISSUE #679: bugged line at the bottom of the viewport
most obvious when targeting or aiming. on latest aurora
Image

dooplecks:
Fixes #679.

There doesn't seem to be anything drawn after the trim that relies on the scissor it sets, except for `calcFade` (which ends up overriding scissor on `TARGET_PC` anyway?).

### DIFF
--- a/src/m_Do/m_Do_graphic.cpp
+++ b/src/m_Do/m_Do_graphic.cpp
@@ -1237,8 +1237,10 @@ static void trimming(view_class* param_0, view_port_class* param_1) {
 
         GXEnd();
     }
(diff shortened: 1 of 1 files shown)


encounter:
Can we add a comment explaining why we commented it out

encounter:
Is this resolved by https://github.com/encounter/aurora/pull/157 or is this a separate issue?

dooplecks:
Separate issue. I don't think there's a real fix for this at the Aurora level anyway.

PR #685, opened 2026-05-05
Recording Mode

MelonSpeedruns:
Needs to be discussed. I think having it on main would help for making trailers in the future without having to maintain this branch.


encounter:
One thing to consider is that "Disable Main HUD" is something experienced people may be comfortable playing with, while "Recording Mode" probably hides too much for that purpose.

encounter:
Let's at least move this to Interface?

MelonSpeedruns:
> Let's at least move this to Interface?

Done!

PR #687, opened 2026-05-05
Fix THP volume being too low

MelonSpeedruns:


encounter:
daMP_THPPlayerSetVolume internally clamps the value to 127, so this just makes >=50% mean 100%. Will probably want a better solution

PR #688, opened 2026-05-05
Change default volume to 50% instead of 80%

MelonSpeedruns:


encounter:
Will want Maddie input on the volume stuff first

qwertyquerty:
You should change the volume slider to be exponential instead, as 50% is basically -3db with the current implementation

In a proper implementation 100% would be -0db and 0% would be -infinity db and every 10% would be -3db or so, maybe -6, you should mess around with what sounds right to you

encounter:
(To be clear, we should change the thing that reads the volume value, not the slider itself)

qwertyquerty:
I'm reimplementing this in #1168

PR #698, opened 2026-05-07
Android fixes

CraftyBoss:
Android now can persist ISO file permission across app reloads, instead of needing to re-select iso every load. Also added resource folder to android build system so rmlui works properly now. Also added a new android-specific ImGui file access so SDL I/O is used instead of the default c api.


encounter:
Could we make the assets/res dir a symlink instead of copying it

CraftyBoss:
Is that something that would work cross platform? the other option could be to move the root res folder into a subfolder, which i could then point gradle to

src/dusk/ui/prelaunch.hpp
```diff
@@ -36,6 +36,9 @@ class PrelaunchOptions;
 
 struct PrelaunchState {
     bool initialized = false;
+#if TARGET_ANDROID
+    bool isIsoPermitted = false;
```
encounter: What is this for

src/dusk/ui/prelaunch.cpp
```diff
@@ -90,6 +93,13 @@ void file_dialog_callback(void*, const char* path, const char* error) {
     state.errorString.clear();
     getSettings().backend.isoPath.setValue(state.selectedDiscPath);
     config::Save();
+
+#if TARGET_ANDROID
+    // store path permissions
+    android::takeUriPermissions(state.selectedDiscPath);
```
encounter: Could we somehow do this on file selection instead of here? It calls into Java for the file dialog stuff, right?

src/dusk/android/ImGuiAndroidFileAccess.cpp
encounter, line 1: This looks like something we can actually upstream to aurora. We generally have been making file access go through SDL there. Nothing about this is specific to Android

encounter:
> Is that something that would work cross platform?

I see no reason why not

src/dusk/ui/prelaunch.hpp
CraftyBoss: verifying if the iso path still has permissions, bc users can remove the permission manually which can cause crashes on boot

src/dusk/android/ImGuiAndroidFileAccess.cpp
CraftyBoss: i can definitely remove it here and move to aurora, i had it here for the sake of this PR, and potentially moving to aurora

src/dusk/ui/prelaunch.hpp
encounter: It doesn't seem to be read, unless I'm missing something

src/dusk/ui/prelaunch.cpp
CraftyBoss: I figured having the option to not always take file permissions for selected files would be nice, but it would be significantly simpler to do it all on the java side, and leave permission verification here.

encounter: Do we need to explicitly check permissions or is trying to open the file enough? I think we're already handling gracefully if we can't open the disc image

CraftyBoss:
> > Is that something that would work cross platform?
> 
> I see no reason why not

this is the part where i expose the fact that i've never made a symlink before

src/dusk/ui/prelaunch.cpp
CraftyBoss: nope android is super cool and throws an error if there's no permission to open a content URI, so a separate check needs to be done prior to attempting to open, though i suppose that could be done thru the activity as well instead of here

encounter:
I figured out a non-symlink approach for res, upstreamed the ImGui SDL_IOStream changes, and fixed a gajillion other things in 1c85ee63ebb3f08f2992dda709509e4b9ecee0b1

PR #700, opened 2026-05-07
Add FPS counter to RmlUi

### LINKED ISSUE #632: New UI: FPS overlay
People probably will want one that's not the ugly F3 debug overlay

Irastris:
Adds a framerate counter to the Graphics tab (renamed to "Video" as I feel that fits better) that can be displayed at a corner of the user's choosing. Closes https://github.com/TwilitRealm/dusk/issues/632

image


encounter:
Maybe combine the 2 options into one, with an Off option?

encounter:
Oh and we should definitely remove the background blur from this component, I want it to be very lightweight

Irastris:
Done, refer to edited OP for a new preview screenshot

PR #709, opened 2026-05-07
Fix OSReport strings corruption with non-trivial format

Antidote:
Use a different method to format the string, this works more reliably and allows OSReport calls to be forwarded to the log.

### DIFF
--- a/src/dusk/OSReport.cpp
+++ b/src/dusk/OSReport.cpp
@@ -25,10 +25,11 @@ static bool checkEnabled() {
 #define va_copy(d, s) ((d) = (s))
 #endif
 
-static std::string FormatToString(const char* msg, va_list list) {    
+static std::string FormatToString(const char* msg, va_list list) {
     size_t size = (strlen(msg) * 2) + 50;
     std::string str;
     va_list ap;
+    int attempts = 0;
     while (true) {
         str.resize(size);
         va_copy(ap, list);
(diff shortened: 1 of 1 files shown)


PJB3005:
Could you explain what's wrong with the previous version?

Antidote:
In the previous version it wouldn't create a buffer big enough on the first pass and return either a truncated or null string.

By being aggressive with the buffer size and then resizing to the actual size after successfully formatting the string we avoid that.

It generally works after 2 passes, but can take as many as 3.

PJB3005:
Does that mean `vsnprintf` doesn't return the same count between calls with the same format string/args?

Antidote:
That appears to be the case, as annoying as it is, par for the course with these string functions unfortunately.

PJB3005:
Why must C be like this...

Antidote:
Your guess is as good as mine, lol

src/dusk/OSReport.cpp
```diff
@@ -21,16 +21,30 @@ static bool checkEnabled() {
...
-        return {};
+#ifndef va_copy
+#define va_copy(d, s) ((d) = (s))
+#endif
+
+static std::string FormatToString(const char* msg, va_list list) {    
+    size_t size = (strlen(msg) * 2) + 50;
+    std::string str;
+    va_list ap;
+    while (true) {
```
PJB3005, line 33: Do we want to have an upper bound on this loop? If `vsnprintf` keeps returning `-1` for a reason unrelated to the input size we'll just keep spinning indefinitely here until OOM.

Frankly I'd say we shouldn't be looping at all if `-1` gets hit but I figure you ran into that in practice? God what a nightmare.

Antidote: Fair point, I'll update it to return an empty string after the 3rd failed attempt, ideally it should get it by then.

Also, unfortunately it can return -1 even if the format itself didn't fail, which is dumb.

Antidote:
Added the attempt counter, and we have OSReport again.

PR #717, opened 2026-05-08
Set max chain links drawn to 600

Pheenoh:
Better fix for the super clawshot crash, just limit the max amount of chain links drawn instead of skipping the actor's draw all together since that skips the ball and chain chain from being drawn as well when super clawshot is on.

600 number comes from the unit max for Argorok divided by the rough # of units per link (3000 / 5)


encounter:
let's remove the option text that was added

PR #721, opened 2026-05-08
Refactor notification settings, allow disabling controller toasts

Irastris:
Replaces "Achievement Notifications" in the Interface tab with a multi-select "Toasts" option, allowing users to disable their choice of toast types.

image


encounter:
Let's call it "Notifications"

PR #722, opened 2026-05-08
Fix Letter Menu Page Numbers

SuperDude88:
See commit message for explanation of the bug. It fixes the issue on my machine, though I'm not sure if this technically forces things to the right order per the standard? Or if a sufficiently aggressive attempt to optimize things might still make it break

This does not appear to affect any other uses of `getBounds`. If possible it would be good to upstream the change to decomp, but I don't have it set up so I can't check if this still matches.


PJB3005:
> though I'm not sure if this technically forces things to the right order per the standard? Or if a sufficiently aggressive attempt to optimize things might still make it break

AFAICT from reading the standard, this should be fine.

PR #741, opened 2026-05-09
Navigate Carousel Without Focus

### LINKED ISSUE #707: Moving controller stick should re-select a UI button after restoring window focus
On some menu screens (internal resolution selection for example), clicking back onto the window will de-select all of the UI buttons. This is fine if you're using a mouse, but if you try to use a controller you can't change the resolution without backing out and re-entering the option (or being careful to click only on the buttons/window title bar).

It would be helpful to have some kind of visual feedback or (preferably) select the button again if you try to move it with a controller and nothing is selected.

SuperDude88:
Allow left/right inputs to change the setting value without having the arrow explicitly clicked. Fixes #707

### DIFF
--- a/src/dusk/ui/graphics_tuner.cpp
+++ b/src/dusk/ui/graphics_tuner.cpp
@@ -208,7 +208,7 @@ GraphicsTuner::GraphicsTuner(GraphicsTunerProps props, bool prelaunch)
         description->SetInnerRML(escape(props.helpText));
     }
     if (auto* carouselParent = mDocument->GetElementById("carousel-container")) {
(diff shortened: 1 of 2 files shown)


src/dusk/ui/graphics_tuner.cpp
```diff
@@ -283,6 +283,10 @@ bool GraphicsTuner::handle_nav_command(Rml::Event& event, NavCommand cmd) {
         return true;
     }
 
+    if(mCarousel && mCarousel->handle_nav_command(cmd)) {
```
encounter: clang-format -i src/dusk/ui/graphics_tuner.cpp

PR #753, opened 2026-05-09
Readme: Add hashes for .RVZ format of GC ROMS

Heliumrich:
Updated the README to include SHA-1 hashes for .RVZ format, USA and EU

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ First, make sure your dump of the game is clean and supported by Dusk. You can d
 
 # Building
 
-If you'd like to build Dusk from source, please read th`e [build instructions](docs/building.md).
+If you'd like to build Dusk from source, please read the [build instructions](docs/building.md).
 
 Pull requests are welcomed! Note that we do not accept contributions that are primarily AI-generated and will close your PR if we suspect as much.
 
--- a/README.md
+++ b/README.md
@@ -23,10 +23,10 @@ It aims to be as accurate as possible to the original while also providing new o
 
 First, make sure your dump of the game is clean and supported by Dusk. You can do this by checking the SHA-1 hash of your dump against this list of supported versions:
 
-| Version      | SHA-1 hash                                 |
-|--------------| ------------------------------------------ |
-| GameCube USA | `75edd3ddff41f125d1b4ce1a40378f1b565519e7` |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -50,7 +50,7 @@ First, make sure your dump of the game is clean and supported by Dusk. You can d
 
 # Building
 
-If you'd like to build Dusk from source, please read the [build instructions](docs/building.md).
+If you'd like to build Dusk from source, please read th`e [build instructions](docs/building.md).
```
Antidote: Have a stray grave here

Heliumrich: Woops, I did this on my smartphone lol, thanks

PJB3005 (changes requested):
Uh, isn't RVZ a compressed format? That can't have consistent hashes can it?

Heliumrich:
> Uh, isn't RVZ a compressed format? That can't have consistent hashes can it?

It is a compressed format, but you think it would have some randomness ? That would be weird

It is officialy supported by Dolphin, and you can go back to a .ISO

PJB3005:
> It is a compressed format, but you think it would have some randomness ?

The file produced by the compression can be completely different depending on compression settings or version of compression library used.

The hash you are putting here (I assume) is the hash of the `.rvz` itself rather than the *compressed contents* of it.

Heliumrich:
> > It is a compressed format, but you think it would have some randomness ?
> 
> The file produced by the compression can be completely different depending on compression settings or version of compression library used.
> 
> The hash you are putting here (I assume) is the hash of the `.rvz` itself rather than the _compressed contents_ of it.

Hmm, that's using latest dolphin, and default compression setting.

I don't think the library is getting improved much nowaday, and most people would keep the defaults settings, but yes I guess you are right

I just did a SHA-1 hash of the RVZs themselves yes.

PJB3005:
> I don't think the library is getting improved much nowaday

zstd receives regular updates still, actually.

Heliumrich:
Why not add a feature to make the app check the hash when we select a ROM ?

As I understand, the app is decompressing the RVZ anyway, so it should be able to get the .ISO hash

encounter:
Yeah, every RVZ will be a different hash, depending on if it was made by Dolphin, NKit 2, nod, or something else. Dusk will verify RVZ files properly when you select one though, so no need for hashes. Thanks though!

PR #782, opened 2026-05-10
Prevent zipping of appimage

Dashboy1998:
Changing the workflow so the appimage is not zipped as [that is explicitly mentioned in the docs.](https://docs.appimage.org/packaging-guide/distribution.html#do-not-put-appimages-into-other-archives)

It would also allow users to get auto updates using [Gearlever](https://github.com/mijorus/gearlever/issues/446)


PJB3005 (approved):

encounter:
This isn't where the zipping happens for release, but I will remove it

PR #814, opened 2026-05-10
Allow menu navigation via D-Pad

RibShark:
Updates d_lib.cpp to accept D-Pad inputs alongside left stick. Left stick is prioritized if it is in use (stickValue != 0), otherwise D-Pad input is checked.

### DIFF
--- a/src/d/d_lib.cpp
+++ b/src/d/d_lib.cpp
@@ -152,7 +152,7 @@ u8 STControl::checkTrigger() {
                 field_0x22 = -field_0x24;
             }
         }
-
+#if !TARGET_PC
         if (!(mDirectionTrig & 3)) {
             Xinit();
         }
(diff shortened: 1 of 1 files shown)


PJB3005 (dismissed):
Please put your changes behind `#if TARGET_PC` checks.

RibShark:
> Please put your changes behind `#if TARGET_PC` checks.

Should I warp the entire thing in a check or just line 157->168 since the logic (mDirectionTrig will be zero always, so the if statements around xinit/yinit will always be true) ends up being the same (and would be optimized out by the compiler most likely)?

MelonSpeedruns:
The code you removed should be re-added between `#if !TARGET_PC` and the code you added between `#if TARGET_PC`

Basically so if TARGET_PC is not defined, then the original code runs ^^

RibShark:
Done!

src/d/d_lib.cpp
```diff
@@ -152,18 +152,43 @@ u8 STControl::checkTrigger() {
...
+    }
+
+    if (!(mDirectionTrig & 3)) {
+        Xinit();
+    }
+
+    if (!(mDirectionTrig & 0xC)) {
+        Yinit();
+#endif
     }
```
encounter, line 192: endif needs to be moved down one line

Irastris (approved):

PR #870, opened 2026-05-10
Option to Invert First Person Aiming

SuperDude88:
Would be helpful if some others could stress test this to make sure it behaves properly in all scenarios/control schemes.

Also included inverted Y aiming (so up on the stick aims up) in the Dusk preset since it's what TPHD (and WWHD/others) do, but if that's undesired I can remove it.

### DIFF
--- a/include/dusk/settings.h
+++ b/include/dusk/settings.h
@@ -144,6 +144,8 @@ struct UserSettings {
         ConfigVar<bool> freeCamera;
         ConfigVar<bool> invertCameraXAxis;
         ConfigVar<bool> invertCameraYAxis;
+        ConfigVar<bool> invertFirstPersonXAxis;
+        ConfigVar<bool> invertFirstPersonYAxis;
         ConfigVar<float> freeCameraSensitivity;
         ConfigVar<bool> debugFlyCam;
         ConfigVar<bool> debugFlyCamLockEvents;
--- a/src/d/actor/d_a_alink_link.inc
+++ b/src/d/actor/d_a_alink_link.inc
@@ -119,8 +119,8 @@ BOOL daAlink_c::setBodyAngleToCamera() {
                 var_f31 /= dComIfGp_getCameraZoomScale(field_0x317c);
             }
 
-            shape_angle.y = shape_angle.y + (var_f31 * cM_ssin(mStickAngle));
-            sp8 = mBodyAngle.x + (var_f31 * cM_scos(mStickAngle));
+            shape_angle.y = shape_angle.y + (var_f31 * cM_ssin(mStickAngle) IF_DUSK(* (dusk::getSettings().game.invertFirstPersonXAxis ? -1.0f : 1.0f)));
+            sp8 = mBodyAngle.x + (var_f31 * cM_scos(mStickAngle) IF_DUSK(* (dusk::getSettings().game.invertFirstPersonYAxis ? -1.0f : 1.0f)));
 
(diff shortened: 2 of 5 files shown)


gymnast86:
Works well enough for me but it also seems to invert the gyro aiming as well. There are already separate gyro invert options though (invert pitch and invert yaw) so I don't think this should also apply to gyro aiming

SuperDude88:
> Works well enough for me but it also seems to invert the gyro aiming as well. There are already separate gyro invert options though (invert pitch and invert yaw) so I don't think this should also apply to gyro aiming

My only hesitation with leaving gyro separate is that if someone toggles it during gameplay they might be surprised if it only changes sticks. Curious if there's more opinions on this, if not I'll separate the two.

AL2009man:
> > Works well enough for me but it also seems to invert the gyro aiming as well. There are already separate gyro invert options though (invert pitch and invert yaw) so I don't think this should also apply to gyro aiming
> 
> My only hesitation with leaving gyro separate is that if someone toggles it during gameplay they might be surprised if it only changes sticks. Curious if there's more opinions on this, if not I'll separate the two.

As a person who isn't used to Inverted Camera: tying it to both joystick and gyro is not a great idea. I won't be able to use Normal Camera because the Gyro in first-person mode will invert when I don't want it that way. 

As a result: I'm forced to stick with "Invert Joystick Camera" while using Gyro.

SuperDude88:
> > > Works well enough for me but it also seems to invert the gyro aiming as well. There are already separate gyro invert options though (invert pitch and invert yaw) so I don't think this should also apply to gyro aiming
> > 
> > 
> > My only hesitation with leaving gyro separate is that if someone toggles it during gameplay they might be surprised if it only changes sticks. Curious if there's more opinions on this, if not I'll separate the two.
> 
> As a person who isn't used to Inverted Camera: tying it to both joystick and gyro is not a great idea. I won't be able to use Normal Camera because the Gyro in first-person mode will invert when I don't want it that way.
> 
> Which means: I'm forced to stick with "Invert Joystick Camera".

You are still able to invert the gyro separately on the input page if just the gyro is flipped, it was mostly a question of which arrangement is more intuitive/obvious. Either way it is possible to switch them independently.

AL2009man:
> > > > Works well enough for me but it also seems to invert the gyro aiming as well. There are already separate gyro invert options though (invert pitch and invert yaw) so I don't think this should also apply to gyro aiming
> > > 
> > > 
> > > My only hesitation with leaving gyro separate is that if someone toggles it during gameplay they might be surprised if it only changes sticks. Curious if there's more opinions on this, if not I'll separate the two.
> > 
> > 
> > As a person who isn't used to Inverted Camera: tying it to both joystick and gyro is not a great idea. I won't be able to use Normal Camera because the Gyro in first-person mode will invert when I don't want it that way.
> > Which means: I'm forced to stick with "Invert Joystick Camera".
> 
> You are still able to invert the gyro separately on the input page if just the gyro is flipped, it was mostly a question of which arrangement is more intuitive/obvious. Either way it is possible to switch them independently.

I just tested it. that solves my issue. but when it comes to the actual question itself: Personally, I think it might bring a tiny bit of confusion if turning on the FPS Invert retroactively makes the gyro inverted, solving it by setting Gyro Y. that's one extra option i need to toggle specifically to Gyro..

Personally, I think the gyro portion should be separated from Invert First Person Aiming like most games with native gyro already does. 

edit: darn, i forgot about Gyro Rollball, i didn't accounted for that. 😔

PR #875, opened 2026-05-10
Flake: Desktop Icon & Version String

Krutonium:
Screenshot From 2026-05-10 14-14-11
image

This PR modifies Flake.nix to 
 - Create a .desktop icon when installed on a Nix system
 - Set the version string to `NixOS-shortCommitHash` or `NixOS-dirty` if the tree is dirty.

The version string is up for debate; but I figured it was a reasonable middle ground. By the nature of how flakes work; unless I went and hardcoded the version number to be a specific value, which wouldn't be accurate unless pinned to a specific revision, it just otherwise reads UNKNOWN-VERSION

With this, you know the user is on NixOS, and the specific commit that they're playing on.

Edit: It also adds libdecor to the build so that dusk has decorations on Wayland/Gnome.


flake.nix
```diff
@@ -66,11 +69,23 @@
           "-DFETCHCONTENT_SOURCE_DIR_IMGUI=${imgui-src}"
           "-DFETCHCONTENT_SOURCE_DIR_RMLUI=${rmlui-src}"
           "-DCMAKE_CROSSCOMPILING=ON" # Tests are not working as I didn't want to work through getting google's test suite working as well. This is the only guard I could find to disable it.
+          "-DDUSK_WC_DESCRIBE=NixOS-${builtins.substring 0 7 (self.rev or "dirty")}"
+          "-DDUSK_VERSION_STRING=NixOS-${builtins.substring 0 7 (self.rev or "dirty")}"
```
jtrees: I don't think there is anything NixOS specific here, is there? I'm planning to install it on Fedora with home-manager.

Krutonium: I only went with NixOS because it's being built with Nix. I suppose it'd work on Fedora, you might need to wrap it with nixGL? But that's a you issue more than an upstream (here) issue.

Any suggestions for alternatives to NixOS as the string? The goal is to make it clear if someone opens a bug report with one of these builds, what the providence is.

Krutonium:
For anyone who doesn't know Nix; you're allowed to chime in too - is `NixOS` appropriate or should I figure out some other way to format the version string? Reasoning given can be found above.

(Mention in the PR below was by accident, I copied the wrong number)

flake.nix
jtrees: nix

encounter:
I'd prefer nix- instead of NixOS- in the version string. Also needs a conflict resolved

flake.nix
```diff
@@ -47,6 +47,9 @@
           cp -r ${aurora-src}/. $sourceRoot/extern/aurora/
           chmod -R u+w $sourceRoot/extern/aurora
           sed -i '/add_subdirectory(tests)/d' $sourceRoot/extern/aurora/CMakeLists.txt
+          # Allow Setting VERSION
+          sed -i 's/find_package(Git)/# find_package(Git)/' $sourceRoot/CMakeLists.txt
+          sed -i 's/set(DUSK_WC_DESCRIBE "UNKNOWN-VERSION")/# set(DUSK_WC_DESCRIBE "UNKNOWN-VERSION")/' $sourceRoot/CMakeLists.txt
```
encounter: You could wrap the version stuff in CMakeLists.txt in `if (NOT DEFINED DUSK_WC_DESCRIBE)` or similar instead of patching it

Krutonium: Noted!

Krutonium:
image
image

encounter:
Thanks!

PR #879, opened 2026-05-10
Crash Reporting Popup

SuperDude88:
If the build has crash reporting, adds a menu on startup to explain what is included/give you the option to disable it. Does not handle reports before the UI is totally set up, those cases should be handled via the cvar command line option if someone is really concerned about it.


encounter:
I'm not necessarily against this, but from a UX perspective I think we already have enough things pop up at startup, so I'm iffy

PR #931, opened 2026-05-11
RMLUI Localization

sera-phie:
Some simple JSON string storage for eventual localization


CMakeLists.txt
```diff
@@ -405,6 +405,7 @@ set(GAME_BASE_FILES
         ${REL_FILES}
         ${DUSK_FILES}
         ${DOLPHIN_FILES}
+        "src/dusk/ui/lang.cpp"
```
qwertyquerty, line 418: This should go in files.cmake

qwertyquerty:
We should look into using RMLUI TranslateString

https://mikke89.github.io/RmlUiDoc/pages/cpp_manual/interfaces/system.html

for this

qwertyquerty:
We already depend on nlohmann/json so you should use that for any parsing, if we do end up storing localization in JSON

Issue #942, opened 2026-05-11
(Linux) Fishing log always displays record as 5 inches


CirkuitBreaker:
<img width="1600" height="1200" alt="Image" src="https://github.com/user-attachments/assets/b9f658ac-605b-49c4-81f9-940b5021cc35" />

tomlube:
What hardware are you on? Can't replicate on windows.

CirkuitBreaker:
OS: Nobara Linux 43 (like Fedora)
Kernel: Linux 7.0.5-200.nobara.fc43.x86_64
CPU: AMD
GPU: AMD 7900 XTX
AppImage Version: 1.0.1

CirkuitBreaker:
The save file appears to be correctly keeping a record of the largest fish of each type, as I only get the text/jingle when catching an actual new record. The fishing journal just displays the incorrect length.

CirkuitBreaker:
[01-GZ2E-gczelda2.gci.bak.gz](https://github.com/user-attachments/files/27579655/01-GZ2E-gczelda2.gci.bak.gz) Providing save file.

PR #948, opened 2026-05-11
Only set isEnglish for GCN PAL English

Pheenoh:
`isEnglish` is used for determining the unit conversion, which isn't correct for GCN USA (we use freedom units by default)


Pheenoh:
Closes #942

PJB3005 (approved):

PR #950, opened 2026-05-11
Fix Prelaunch Break-Out With Controller Config

### LINKED ISSUE #945: Softlock on pre-launch screen
Reproduction steps:
1) Open Dusk and go into pre-launch settings instead of selecting "Play."
2) Navigate to Input > Configure Controller.
3) Press the View button.
4) The settings window is dismissed and cannot be re-opened. You can press the Start button, and the Twilight Princess logo will load. However, attempting to continue after this point results in an endless loop of "No memory card found in Slot A. Continue without saving?" because "Yes" cannot be selected.

Tested on Dusk v1.0.1, Windows 10.

SuperDude88:
Fixes #945

### DIFF
--- a/src/dusk/ui/controller_config.cpp
+++ b/src/dusk/ui/controller_config.cpp
@@ -301,7 +301,11 @@ int rumble_raw_to_percent(u16 raw) {
 
 }  // namespace
 
-ControllerConfigWindow::ControllerConfigWindow() {
+ControllerConfigWindow::ControllerConfigWindow(bool prelaunch) {
+    if(prelaunch) {
+        mSuppressNavFallback = true;
+    }
+
     listen(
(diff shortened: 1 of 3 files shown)


src/dusk/ui/controller_config.cpp
```diff
@@ -301,7 +301,11 @@ int rumble_raw_to_percent(u16 raw) {
 
 }  // namespace
 
-ControllerConfigWindow::ControllerConfigWindow() {
+ControllerConfigWindow::ControllerConfigWindow(bool prelaunch) {
+    if(prelaunch) {
```
encounter:
```suggestion
    if (prelaunch) {
```

encounter (approved):

PR #1011, opened 2026-05-11
Add mouse input option for the third-person camera

### LINKED ISSUE #766: Option to use Mouse Input for Free Look
Noticed Free Look doesn't work with Mouse Input as you cannot bind the Mouse Movement to the C-Stick.  
Would make for a better Mouse & Keyboard experience.

### LINKED ISSUE #1685: Feature Request | Hide Mouse Cursor option
I'm using SteamOS on a Go 2, and with that I need to use the mouse as gyro, on opening the program the mouse cursor stays visible and with Gyro it moves with it, it's a bit obnoxious, it can be hidden if I open and close the Dusklight settings menu, there should be an option to hide mouse cursor

Requested : 

- Add a "Hide cursor" option

Irastris:
I've added the ability to control the third-person camera with mouse input sources, including Steam Deck and Steam Controller. Together with mouse aiming, I believe this should make the vast majority of the game playable with a mouse and keyboard.

Most testing was done on my Windows PC with a Logitech G Pro running at 800dpi. I played through the entirety of Arbitor's Grounds without any issues. Some additional testing sessions were performed on a Steam Deck to ensure mouse and joystick inputs could coexist, and that worked as well.

The only known missing camera movement is when L-targeting, you can not move the camera left and right with the mouse. I did not personally have a problem with this, but it's something to consider supporting in the future if possible.


Irastris:
Progress as of May 22nd, all testing done on Windows:

Mouse input for the third-person camera is working and feels great. The mouse is captured so that it cannot leave the window boundaries, and released as soon as window focus is lost, a RmlUi document is brought up, or the mouse camera is disabled. I spent an hour on the 20th playing and encountered no major issues, tidied the code up afterward, and will continue to play as I work on this PRs feature-set.

There is an outstanding TODO where I disabled the ability for the camera to be controlled by both a joystick and mouse at the same time. It's probably nothing that anyone would actually want, but I'd still like that to be functional.

Logic is completely intertwined with Free Camera at the moment, which means it shares any shortcomings that may have. At this time, the only problem I've found to resolve is that there is (mostly) no camera control when on horseback.

Note to self, do not forget about existing cursor visibility logic currently in ImGuiConsole.

Azazel0671:
I compiled and tested on linux and the only two issues I found were the horse thing and also when targeting an enemy you cant look left and right like you normally would be able to with the C stick.

missingnoXXVI:
Keyboard usage still leaves plenty to be desired atm, but outside of having to reduce all of my sensitivities down to 27% percent because of being, well, quite sensitive and the already mentioned issues, this does work great. Can't help but hope this play gets imported into other Zelda PC releases. Love my M&K.

missingnoXXVI:
👍👍 on Epona. Once combat and keyboard get more love, I will consider this the first unofficial PC release I have played with proper M&K support.

missingnoXXVI:
Just tried the latest and now the sensitivity percentage doesn't seem to be effecting regardless of being at 25% or 400%

missingnoXXVI:
Couldn't test the other 2 action builds, but the one with the sensitivity changes fixed the issue. 400% spins out like absolute crazy as one would expect it to and 25% is a matter of preference 👍👍

Azazel0671:
Are there plans to make the mouse able to move the camera while z targeting the same way the c stick can? (at least side to side the same way it behaves)

Irastris:
Not in this PR, maybe in the future.

Issue #1012, opened 2026-05-11
Hitching a ride up waterfall with Zora at Zora's Domain crashes the game.


Harpig:
On Steam Deck
Dusk v1.0.1
There's a Zora at the base of the waterfall in Zora's Domain, whom you can ask to take you up the waterfall to the throne room. I've tried this twice and both times crashed the game.

tomlube:
Can you post a log? Feel free to open it and remove your username/path

Harpig:
[dusk-20260511-175628.log](https://github.com/user-attachments/files/27606012/dusk-20260511-175628.log)

Krutonium:
[dusk-20260511-154734.log](https://github.com/user-attachments/files/27609465/dusk-20260511-154734.log) Here's a second log of the crash. I don't immediately see anything relevant. When running it in a terminal though; the output at the end that's not captured in the log is
```
*** buffer overflow detected ***: terminated
fish: Job 1, 'dusk' terminated by signal SIGABRT (Abort)
```

Desmondchaser27:
Tested this as well on CachyOS/Arch.  This still crashes on version 1.1.1 release.

Harpig:
Can confirm the cutscene now works on v1.1.1
Hesistant to close if others ( @Desmondchaser27 ) are still experiencing issues.

PJB3005:
Closing unless other people show up with the same crash yeah.

Desmondchaser27:
So yes, I did just test again on 1.1.1 and it does crash consistently.  Here is the log file.  Log file is super simple.  Just opened the game where I saved before the waterfall.  Then talked to the Zora and had the crash happen. 

[dusklight-20260514-120809.log](https://github.com/user-attachments/files/27767782/dusklight-20260514-120809.log)

iobradovi:
hi, yeah the game freezes and then just closes if i say yes when the zora at the waterfall asks me whether i want to get to the throne room. nothing shows up in the logs. cachyos kde

[dusklight-20260515-220404.log](https://github.com/user-attachments/files/27816887/dusklight-20260515-220404.log)

l0ud:
Still crashes on linux on v1.2.0.

tomlube:
> Still crashes on linux on v1.2.0.

only at night, right?

iobradovi:
can this issue be reopened? lots of people have reported having the same issue and there was a duplicate too and the issue is still present

Abzol:
[...]

PR #1042, opened 2026-05-11
Fix Overflow/Off-by-one. Fixes #1036 and #1012

### LINKED ISSUE #1036: Putting the sword in the pedistal in the Temple of Time crashes the game
Image

[dusk-20260511-175144.log](https://github.com/user-attachments/files/27612628/dusk-20260511-175144.log)

At the end of the console output, it prints:

*** buffer overflow detected ***: terminated
fish: Job 1, 'dusk' terminated by signal SIGABRT (Abort)

Which matches #1012 which may be related.

This represents a softlock as you are unable to progress the game past this point.

(Verified on AppImage 1.0.1 as well)

Krutonium:
As seen in #1036 and #1012 20 is simply not enough space for some event names. Unsure if there's yet longer ones, but the length of the event names is 20 + Terminator, or 21.

### DIFF
--- a/include/d/d_event.h
+++ b/include/d/d_event.h
@@ -196,7 +196,7 @@ public:
     /* 0x108 */ int mSkipTimer;
     /* 0x10C */ int mSkipParameter;
     /* 0x110 */ BOOL mIsSkipFade;
-    /* 0x114 */ char mSkipEventName[20];
+    /* 0x114 */ char mSkipEventName[21];
     /* 0x128 */ u8 mCompulsory;
     /* 0x129 */ bool mRoomInfoSet;
     /* 0x12C */ int mRoomNo;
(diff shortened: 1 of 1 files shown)


PJB3005 (changes requested):
This needs to be behind `#if TARGET_PC`

Krutonium:
Added the guard; Checking it compiles right now.

Edit: Yep, compiles and works :+1:

PJB3005 (approved):

PJB3005:
This wouldn't be the first off-by-one they had like this so yeah, seems reasonable to me.

MelonSpeedruns:
LGTM

Krutonium:
If you're curious, the string that I ended up catching in the crash was `LV6_GATE_APPEAR_SKIP` - Which fills the entire 20 characters, and then the null terminator gets stuck on the end, and then glibc throws because of the overflow. 

I don't really know C++ that well so it was an adventure playing with GDB to trace this back. But... Results!

PR #1044, opened 2026-05-11
Add Nix devshell for Linux and macOS development

SailorSnoW:
Adds `devShells.<system>.default` for `x86_64-linux`, `aarch64-linux`, `x86_64-darwin`, and `aarch64-darwin` so the repo can be developed via `nix develop` / direnv on both platforms. The existing `packages.x86_64-linux.default` build is unchanged.

- Linux (`mkShell`): gcc + clang/lld, cmake, ninja, pkg-config, python3 + markupsafe, rustc/cargo, sccache, plus the system libs mirrored from the Ubuntu apt list in `.github/workflows/build.yml` (X11/Wayland, Vulkan, GL, ALSA/PulseAudio/PipeWire, GTK3, freetype, zstd, ...).
- macOS (`mkShellNoCC`): same tooling without Nix's cc-wrapper, so CMake picks up Apple Clang and the Xcode SDK directly — matches the `build-apple` CI job.
- Ignores `.direnv/` and `.envrc`.

Verified with `nix flake check --no-build --all-systems` and by entering the macOS arm64 devshell.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,10 @@ compile_commands.json
 # MacOS
 .DS_Store
 
+# direnv / nix
+.direnv/
+.envrc
+
 # ISOs
 *.iso
 
--- a/flake.nix
+++ b/flake.nix
@@ -4,101 +4,212 @@
   };
   outputs = { self, nixpkgs }:
     let
-      pkgs = import nixpkgs { system = "x86_64-linux"; };
-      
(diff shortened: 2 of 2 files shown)


Krutonium:
Independently verified the `x86_64-linux`  dev shell, everything looks to be in order.

LGTM

PR #1069, opened 2026-05-12
Add a setting to disable the faint buzzing noise when a context menu …

YLFAndy:
This change adds a setting to disable the slight 'buzzing' noise when context-aware button prompts display. It simply adds a trigger to the culling workflow IFF the setting is enabled.


YLFAndy:
Also, if this is not implemented to your liking, feel free to close it and I'll open it as a feature request. The sound just has me reaching for my phone constantly and triggers my specific brand of audio aversion.

src/Z2AudioLib/Z2SeMgr.cpp
```diff
@@ -1169,6 +1170,9 @@ bool Z2SeMgr::isSoundCulling(JAISoundID soundID) {
         }
         field_0x3c8 = 1;
         break;
+    case Z2SE_SY_HINT_BUTTON:
+            return (dusk::getSettings().game.noContextAwareSound);
+        break;
```
encounter:
```suggestion
#ifdef TARGET_PC
    case Z2SE_SY_HINT_BUTTON:
        return dusk::getSettings().game.noContextAwareSound;
#endif
```

YLFAndy: Thanks!

qwertyquerty:
I don't really see this being a widely used option; if it's something you want very badly you can always play on your own fork or reimplement the change with the mod api once that's put in a release.

PR #1071, opened 2026-05-12
improved speedrun mode

qwertyquerty:
- move speedrun timer to rmlui
- restyle timer and make it less intrusive
- speedrun mode settings layer overrides
- proper settings overrides decided upon by TPSR community
- add opton to disable rta timer
- add reset timer button to rmlui bar
- auto connect to livesplit at any time and poll for disconnect
- fix some socket bugs


encounter (approved):

src/dusk/ui/settings.cpp
```diff
@@ -172,116 +166,46 @@ AuroraBackend configured_backend() {
...
-    return path.lexically_normal();
-}
-
-std::filesystem::path user_home_path() {
-    const char* homePath = SDL_GetUserFolder(SDL_FOLDER_HOME);
-    if (homePath == nullptr || homePath[0] == '\0') {
-        return {};
-    }
-    return std::filesystem::path{reinterpret_cast<const char8_t*>(homePath)};
+    getSettings().game.enableTurboKeybind.setSpeedrunValue(false);
```
TakaRikka, line 175: should we also turn on the most common ones that will pretty much always be on? i could see people getting annoyed having to manually adjust settings for speedruns instead of just "turn on speedrun mode and good to go"

TakaRikka (commented):
mostly lgtm, also just some conflicts to fix

src/dusk/ui/settings.cpp
qwertyquerty: I dont think so, I think speedrun mode should only touch constrained settings and shouldn't mess with people's existing settings

TakaRikka: fair enough

PR #1095, opened 2026-05-12
Update README.md

Rvervuurt:
The download link was broken


PJB3005:
I'm pretty sure the repo's supposed to be renamed, so this isn't a great fix.

Rvervuurt:
Ah, I wasn't aware how the name-change was going to be, if it was from dusk > dusklight or the other way around. Sure, if it's supposed to be dusklight, the repo has to change names, but for now, the download link is broken.

PR #1103, opened 2026-05-12
feat: auto-activate portable mode when 'portable/' folder exists next to executable

WilliamFriconneau:
## Summary

- Adds `host_dir()`: resolves the directory containing the AppImage file via the \`\$APPIMAGE\` env var (falls back to \`SDL_GetBasePath()\` for native executables — \`SDL_GetBasePath\` points to the temporary mount point inside an AppImage, not the actual file location)
- Adds public `host_path_relative()` in \`data.hpp\`
- Auto-portable detection in \`initialize_data()\` and \`configured_data_path()\` now uses \`host_path_relative\` so the \`portable/\` folder is resolved relative to the AppImage file on disk
- Auto-scans the host directory for disc images (\`.rvz\`/\`.iso\`/\`.gcm\`/\`.wia\`/\`.ciso\`/\`.gcz\`/\`.wbfs\`) at startup when no \`--dvd\` argument is provided; opens the first valid one found

## Behaviour

Place a \`portable/\` folder next to the AppImage → all data is stored there automatically.  
Place the disc image next to the AppImage → it is opened automatically without a file picker.

Both features require no flags, no config, no marker files.

## Motivation

AppImage + portable folder + disc image in the same directory: one folder to copy, everything works out of the box. Consistent with how projects like Eden handle portability.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/dusk/data.cpp
+++ b/src/dusk/data.cpp
@@ -124,14 +124,6 @@ std::filesystem::path base_path_relative(const std::filesystem::path& path) {
     return std::filesystem::path{basePath} / path;
 }
 
-std::filesystem::path host_path_relative(const std::filesystem::path& path) {
-    const auto dir = host_dir();
-    if (dir.empty()) {
-        return {};
-    }
-    return dir / path;
-}
-
 std::filesystem::path default_data_path(const std::filesystem::path& prefPath) {
 #ifdef __APPLE__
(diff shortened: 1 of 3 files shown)


PJB3005:
> Note that we do not accept contributions that are primarily AI-generated and will close your PR if we suspect as much

PR #1109, opened 2026-05-12
MinimalHUD improvement

DarkMicro:
Hello!

It makes more sense to only hide the static button prompts on the main hud so you can still see the important hud elements.
I made a seperate option for the current behavior of Minimal HUD. "Disable HUD"

My solution is to move the button prompts off-screen so the item wheel buttons are still visible. (Because they use some common values like scale and alpha)

I hope you like it.

[Bildschirmaufzeichnung vom 2026-05-12 16-39-32.webm](https://github.com/user-attachments/assets/d5fafcc1-94e6-447a-9f88-1d67f44a31c5)


MelonSpeedruns:
If the mini-map is still toggleable, then this looks good to me!

Wham-Ba:
I'm merely an end-user, and have no relation whatsoever to the development team, so keep in mind that my opinion on this is of little value.

But would it not make more sense to merge these two HUD options into a single one labelled "HUD visibility" which could then be set three different values: "On", "Off", and "Minimal"?

MelonSpeedruns:
> I'm merely an end-user, and have no relation whatsoever to the development team, so keep in mind that my opinion on this is of little value.
> 
> But would it not make more sense to merge these two HUD options into a single one labelled "HUD visibility" which could then be set three different values: "On", "Off", and "Minimal"?

This is actually a good idea that I'm in favor of. I think this PR should implement that.

DarkMicro:
What do you think 1 or 2?
1: Like suggested by @Wham-Ba 
Bildschirmfoto vom 2026-05-13 09-06-06

2: You can disable the elements you want
Bildschirmfoto vom 2026-05-13 12-19-18

MelonSpeedruns:
I love 2! Maybe have a space between All Off and Health? Otherwise if everything works then this is great. Also might need you to change the description to something like `Toggle various elements of the main HUD of the game. Useful for a more immersive experience.`

DarkMicro:
It looks like this now:
Bildschirmfoto vom 2026-05-13 19-22-24

src/d/d_meter2.cpp
```diff
@@ -253,6 +253,37 @@ int dMeter2_c::_create() {
 int dMeter2_c::_execute() {
     JKRHeap* heap = mDoExt_setCurrentHeap(mpHeap);
 
+    if (dusk::getSettings().game.ingameHudMode != dusk::IngameHudMode::On) {
+        dusk::IngameHudMode hudMode = dusk::getSettings().game.ingameHudMode;
+        if (!(static_cast<int>(hudMode) & static_cast<int>(dusk::IngameHudMode::Health))){
+            g_drawHIO.mLifeGaugePosX = 1E5;
```
encounter, line 259: Any particular reason we're using pos instead of alpha?

encounter (commented):
Looks good overall, just curious if alpha is a better choice here

src/d/d_meter2.cpp
DarkMicro: For Health, Action Buttons, D-Pad, Lamp-Meter and Oxygen-Meter the usage of alpha is not possible.
Health: I did not find the global alpha value for the health bar.
Action Buttons: If you make them invisible, you cannot see them in the item wheel.
D-Pad: The alpha value did not work at all.
Lamp-Meter and Oxygen-Meter: Instead of being invisible, they just flicker.

Rupees and keys would work with alpha but it is not necessary to do it different unlike the Vessel of Light.
The Vessel of Light would just move very slowly to its new position if you change it. It would take like 10 Seconds to appear if turn it back on.

include/dusk/settings.h
```diff
@@ -34,6 +34,19 @@ enum class GyroMode : u8 {
     Mouse = 1,
 };
 
+enum class IngameHudMode : int {
+    Off = 0,
```
GiorgioMendieta, line 38: Would it be cleaner to use bit shifting instead like 2 << 1, 2<<2 ... 2<<8 ?

DarkMicro: I don't know. I think it is more complicated syntax with bit shifting. I think everyone knows that 1, 2, 4, 8... are bits and 255 is the sum of all.

GiorgioMendieta:
I think a good QoL improvement for this pull request would be to also hide the Rupee count and instead only show it during the item wheel selector. What do you guys think ?

DarkMicro:
@GiorgioMendieta Cool idea, i look into it

DarkMicro:
@GiorgioMendieta I tried but i don't know where the elements on the main hud and the wheel menu are defined. This seems to be a more complicated change. Maybe this is a good feature for a "Dynamic HUD" option made by a second PR in the future.

PR #1117, opened 2026-05-12
Log build info on startup

PJB3005:


TakaRikka:
should be fine after conflicts resolved

PR #1121, opened 2026-05-12
Allow Mouse mapping

Yiori:
This will allow mouse to be mapped to the c-stick to use the camera


Irastris:
While your contribution is appreciated, I'm afraid we already have different plans for mouse input in the works already (https://github.com/TwilitRealm/dusklight/pull/1011). Thank you anyways!

PR #1123, opened 2026-05-12
Add Invincible Enemies cheat

SailorSnoW:
made this to practice faster S&S Skip/EMS on my side, might be useful for others too so I PR it


encounter:
Thanks for the PR! Looks like we got some merge conflicts to resolve

Issue #1119, opened 2026-05-12
Markers for Statues to be moved not being removed


Krutonium:
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/c4685e9e-beb7-42f3-9fa7-40386b6d25f9" />
<img width="554" height="528" alt="Image" src="https://github.com/user-attachments/assets/b232e750-ba52-43ca-a226-ad606e17f812" />

At this point I've completed "City in the Sky" so well beyond the finishing of them, but the markers that indicate the locations of the statues to move around the world to get the symbols from underneath them have never been removed.

MelonSpeedruns:
This is a vanilla "feature". If you'd like to make an enhancement to make those icons disappear, feel free to make a PR!

Krutonium:
Wait really? Why? Nintendo what are you doing?

PR #1129, opened 2026-05-12
Add ability to toggle off Owl Statue Markers on Map

Krutonium:
I don't have a save file at the correct point in time, but this adds an option under Interface => General that in theory should make it so that once you move the Canon to Lake Hylia, Definitively *after* the markers are needed any more; it'll stop rendering them, cleaning up the map.

I don't *think* this should have any side effects, but I'm largely not a C++ programmer, so if I messed up, please be kind and maybe point at what I did wrong and how I might fix it.

My save is currently post City in the Sky, so when I turn it on, the un-needed map markers disappear immediately.

Inspired by #1119
Screenshot From 2026-05-12 17-44-48
Screenshot From 2026-05-12 17-44-41
Screenshot From 2026-05-12 17-44-20
Screenshot From 2026-05-12 17-43-51
Screenshot From 2026-05-12 17-43-47


MelonSpeedruns (approved):
Looks good to me!

src/d/d_map.cpp
```diff
@@ -931,6 +931,13 @@ bool renderingAmap_c::isDrawIconSingle2(dTres_c::data_s const* i_data, bool para
         }
         break;
     case 5:
+#if TARGET_PC
+        if (dusk::getSettings().game.removeQuestMapMarkers &&
+            dComIfGs_isEventBit(dSv_event_flag_c::saveBitLabels[0x190]))
```
PJB3005, line 936: Don't we have constants for these?

Krutonium: I did a cursory search and didn't see one, but that doesn't mean it doesn't exist.

Krutonium: (I'm also not 100% sure that 0x190 is correct, so uh... I'd appreciate a double check on that!)

MelonSpeedruns:
If it builds on your end just ignore the build checks

Krutonium:
> If it builds on your end just ignore the build checks

It wasn't, but I figured it out. Turns out, and this is a learning experience, C++ really cares about the order of things that I honestly didn't expect to care about having ordered.

PR #1137, opened 2026-05-12
set sdl app metadata when initializing audio

Sulfrix:
The Dusklight SDL audio source appears as "SDL Application" in Pipewire audio mixers.
05-12-26 06 20 22 PM
This PR calls SDL_SetAppMetadata() before initializing the SDL audio source in order to give the audio source a proper name.
image

### DIFF
--- a/src/dusk/audio/DuskAudioSystem.cpp
+++ b/src/dusk/audio/DuskAudioSystem.cpp
@@ -4,6 +4,7 @@
 #include <array>
 #include <cassert>
 #include <span>
+#include <version.h>
 
 #include "JSystem/JAudio2/JASAiCtrl.h"
 #include "JSystem/JAudio2/JASChannel.h"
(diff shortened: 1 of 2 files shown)


PJB3005 (changes requested):
This really isn't the right place to call this. This should be in main or maybe Aurora instead.

Sulfrix:
Ah, the way the audio subsystem was initialized confused me and I assumed the game window itself was not an SDL window. I can move it.

src/dusk/audio/DuskAudioSystem.cpp
```diff
@@ -45,6 +46,7 @@ static int RenderNewAudioFrame();
 static void RenderAudioSubframe();
 
 static void InitSDL3Output() {
+    SDL_SetAppMetadata("Dusklight", DUSK_VERSION_STRING, "dev.twilitrealm.dusklight");
```
Irastris: The application ID is still dev.twilitrealm.dusk, that part has not changed to Dusklight

Sulfrix:
Moved the SDL metadata set to right after the gamepad mappings are loaded.

PR #1150, opened 2026-05-13
Add support for portable mode in Dusk

snnh:
## Title

Add portable data folder startup support

## Summary

Adds startup-level portable mode support so Dusklight can keep user data next to the executable.

- Add a `--portable` command-line option.
- Add `portable.txt` marker support next to the executable.
- Use a sibling `data` folder for settings, saves, logs, and cache in portable mode.
- Avoid touching the platform preference path when portable mode is selected at startup.
- Document portable usage in `docs/building.md`.

## Notes

This is separate from the existing Data Folder / Portable Mode setting UI. The startup flag and marker file are intended for packaged portable builds or users who want portable behavior before preferences are loaded.

This implementation is partially inspired by Dolphin Emulator's portable mode behavior.

## Testing

- Ran `git diff --check`.
- Checked for conflict markers.

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -93,6 +93,18 @@ Alternate presets available:
 #### Running
 Pass the disc image as a positional argument. Supported formats: ISO (GCM), RVZ, WIA, WBFS, CISO, GCZ
 ```sh
-build/{preset}/dusklight/path/to/game.rvz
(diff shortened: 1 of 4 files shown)


encounter:
> Note that we do not accept contributions that are primarily AI-generated and will close your PR if we suspect as much.

PR #1155, opened 2026-05-13
Invert Air/Swim Axes

SuperDude88:
Apparently this is a setting HD has...

Currently only affects swimming, need to look for the flying code (will remain draft until then). Option name is based on HD but I'm open to changing it if we think another wording is more clear, also haven't seen like anyone ask for this so totally fine if it's closed to avoid menu clutter.

### DIFF
--- a/include/dusk/settings.h
+++ b/include/dusk/settings.h
@@ -150,6 +150,8 @@ struct UserSettings {
         ConfigVar<bool> invertCameraYAxis;
         ConfigVar<bool> invertFirstPersonXAxis;
         ConfigVar<bool> invertFirstPersonYAxis;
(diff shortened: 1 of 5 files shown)


encounter:
What would bring this out of draft?

SuperDude88:
> What would bring this out of draft?

I need to find where the controls are handled for flying, then it'll be ready

Abzol:
Instictively they're right here

https://github.com/TwilitRealm/dusklight/blob/7993740ac84a9c951baa9b433f9d1fb78bc83552/src/d/actor/d_a_kago.cpp#L3517

encounter (approved):

PR #1157, opened 2026-05-13
Fixed Debug target not compiling

AceOfSpades-JFK:
Adds an `__MWERKS__` define to the conditional in line 111 of `f_op_actor_mng.h`. This causes compiling to fail because of the classes `fopAc_ac_c` and `leafdraw_class` utilizing the aforementioned define. While the define might've had its use in the discontinued Metrowerks CodeWarrior IDE and the PPC architecture the game was originally made for, it remains unused in the modern platforms running this port.

I added the conditional to keep that code block's original functionality while fixing compilation. With my added fix, the codebase can build fully using the `debug` Cmake target.


include/f_op/f_op_actor_mng.h
```diff
@@ -108,7 +108,7 @@ struct fopAcM_search_prm {
 struct fOpAcm_HIO_entry_c : public mDoHIO_entry_c {
     virtual ~fOpAcm_HIO_entry_c() {}
 
-    #if DEBUG
+    #if DEBUG && __MWERKS__
```
PJB3005: This should be checking `TARGET_PC` IMO.

PJB3005 (changes requested):

include/f_op/f_op_actor_mng.h
AceOfSpades-JFK: Aight. Just pushed a commit with said changes.

PJB3005 (approved):

PR #1172, opened 2026-05-13
ci: publish release assets for tags

banteg:
## Summary
- add a tag-only release job after all platform builds finish
- download workflow artifacts, zip each artifact bundle, and upload them to the matching GitHub Release
- create the release first when the tag exists but no release has been published yet

## Validation
- ruby YAML parse for .github/workflows/build.yml
- git diff --check

Note: this fixes future tag releases. Existing tag v1.1.0 was built successfully, but its workflow came from the old tag commit, so its artifacts still need to be uploaded manually or by retagging/rerunning with this workflow.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -272,3 +272,39 @@ jobs:
             build/install/*.dll
             build/install/res/
             build/install/debug.7z
+
+  release:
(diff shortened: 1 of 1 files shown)


encounter:
We have a separate release script for staging and signing

PR #1212, opened 2026-05-13
Adds option to unbind using controller only

Abzol:
This lets you unbind actions, axes and buttons by inputting the same button or axis again, which should allow mobile device users to unbind keys without having to connect a keyboard

### DIFF
--- a/src/dusk/ui/controller_config.cpp
+++ b/src/dusk/ui/controller_config.cpp
@@ -1013,7 +1013,9 @@ void ControllerConfigWindow::poll_pending_binding() {
         const s32 nativeButton = PADGetNativeButtonPressed(mPendingPort);
         if (nativeButton != -1) {
             const int completedPort = mPendingPort;
-            if (mPendingButtonMapping->nativeButton == static_cast<u32>(nativeButton)) {
+            if (mPendingButtonMapping->nativeButton == static_cast<u32>(nativeButton) and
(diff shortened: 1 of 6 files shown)


Abzol:
The second commit was supposed to go on a branch, whoops, reverted

encounter (approved):

encounter:
Actually, I think we should guard some buttons from getting unmapped (A and B in particualr) so people don't unbind their A button and get stuck

Abzol:
I thought about that case but didn't want to clutter the code with extra magic numbers-style exceptions since every device should have an alternate method of selecting the A button option, but maybe better to write in the safeguard anyway for non-standard setups

Abzol:
I'm sorry for the terrible indentation.

src/dusk/ui/controller_config.cpp
```diff
@@ -1013,6 +1013,12 @@ void ControllerConfigWindow::poll_pending_binding() {
         const s32 nativeButton = PADGetNativeButtonPressed(mPendingPort);
         if (nativeButton != -1) {
             const int completedPort = mPendingPort;
+            if (mPendingButtonMapping->nativeButton == static_cast<u32>(nativeButton) and
```
encounter: sorry to nitpick but let's use &&

Abzol: Completely fair, will check for my other PR as well

PR #1223, opened 2026-05-13
feat: iOS touch controls and gyro

aperezro:
### iOS touch controller support

IMG_0296 2

This PR adds an iOS-specific touch controller path so the game can be played without a physical controller or keyboard.

The implementation adds an on-screen GameCube-style controller overlay for iPhone/iPad, including:

- main control stick
- C-stick
- D-pad
- A / B / X / Y buttons
- L / R / Z buttons
- Start button
- gyro aiming

Touch input is converted into the same `PADStatus` data path used by the existing controller system, so the game receives normal GameCube-style button and stick input.

The touch controller can also be selected in the controller settings menu as **Touch Controls**, alongside the existing Keyboard and physical controller options. When selected, it is treated as an assigned controller for Port 1, which prevents the “No controller assigned” warning from appearing on iOS.

This also adds a touch stick sensitivity option for iOS. The sensitivity value is stored in the app’s user data folder and lets users tune how quickly the virtual stick reaches full tilt.

A separate post-clamp stick application step was added because the normal controller clamp path could reduce virtual stick values back into a walking range. The touch stick now reapplies full post-clamp tilt after the standard PAD clamp, so pushing the virtual stick forward correctly makes Link run instead of walk.

The touch controls are compiled only for iOS/iPadOS and do not affect desktop controller behavior.

The game used to crash on iOS when clicking the keyboard as input, which is now fixed.


PJB3005:
This PR needs to be split up into separate features.

aperezro:
> This PR needs to be split up into separate features.

I removed all the texture related changes since that workaround was RAM inefficient

PR #1237, opened 2026-05-14
Flake: Update GitHub revision and hash for aurora-src

Krutonium:
Fixes Build Error due to Aurora being out of date in the flake.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -15,12 +15,6 @@
 
       # Dependencies that are not packaged in nixpkgs (used by the Linux package build):
       buildSources = pkgs: {
-        aurora-src = pkgs.fetchFromGitHub {
-          owner = "encounter";
-          repo = "aurora";
-          rev = "3643a369ad47d8acca8d73dba462eab2ee2bd150";
-          hash = "sha256-eWLKgj5VNmgF4KH0RFoCEhaPAy0c/8CnVvkuIAsG1AY=";
-        };
         dawn-src = pkgs.fetchzip {
           url = "https://github.com/encounter/dawn-build/releases/download/v20260423.175430/dawn-linux-x86_64.tar.gz";
           hash = "sha256-HXfKTLHtMPwupnFnaflCARtXVPuS/0PoCePXidjE5xs=";
@@ -59,9 +53,6 @@
           name = "dusklight";
           src = ./.;
           postUnpack = ''
-            mkdir -p $sourceRoot/extern/aurora
-            cp -r ${srcs.aurora-src}/. $sourceRoot/extern/aurora/
-            chmod -R u+w $sourceRoot/extern/aurora
             sed -i '/add_subdirectory(tests)/d' $sourceRoot/extern/aurora/CMakeLists.txt
           '';
           # Remove last line to re-enable tests
--- a/flake.nix
+++ b/flake.nix
@@ -18,8 +18,8 @@
         aurora-src = pkgs.fetchFromGitHub {
           owner = "encounter";
           repo = "aurora";
-          rev = "63606a43265a3bc18dafd500ab4d7a2108f109e6";
-          hash = "sha256-xBvnAwGwNzav67Ac6oUz7RqDUwqgL2bsME3OOMn8Tqw=";
+          rev = "3643a369ad47d8acca8d73dba462eab2ee2bd150";
+          hash = "sha256-eWLKgj5VNmgF4KH0RFoCEhaPAy0c/8CnVvkuIAsG1AY=";
         };
         dawn-src = pkgs.fetchzip {
           url = "https://github.com/encounter/dawn-build/releases/download/v20260423.175430/dawn-linux-x86_64.tar.gz";


Krutonium:
Semi Related; Would it be an acceptable idea if I made a PR adding a test that verifies compiling via Nix?

Mastermindaxe:
As a nix flake user of this package I'd love a test that the nix flake compiles, but it's probably too much overhead for the dusklight maintainers if they're not actively using the flake, right?

Mastermindaxe:
Can confirm btw that this pr fixes the flake build. Was able to build it on my machine. Thank you! @Krutonium

voidlily:
> As a nix flake user of this package I'd love a test that the nix flake compiles, but it's probably too much overhead for the dusklight maintainers if they're not actively using the flake, right?

this wouldn't be too hard to do in github actions fwiw

```yaml
...
    steps:
      - uses: actions/checkout@v6
      - uses: cachix/install-nix-action@v31
        with:
          github_access_token: ${{ secrets.GITHUB_TOKEN }}
      - run: nix build
      - run: nix flake check
 ```

(tags should be digests instead of tags for security reasons but this is the rough shape of what it looks like)

**Comment by voidlily** [comment]
Another useful check here would be to make sure the flake inputs match the rest of the project, like aurora-src (submodule), dawnlight, imgui, and so on (urls in cmake files, a few in this repo but also a few in the submodule). Or bigger picture figuring out how to get nix-update-script to do this, thinking ahead to getting this in nixpkgs and making sure it's fully r-ryantm compatible.

**Comment by Krutonium** [comment]
> > As a nix flake user of this package I'd love a test that the nix flake compiles, but it's probably too much overhead for the dusklight maintainers if they're not actively using the flake, right?
> 
> this wouldn't be too hard to do in github actions fwiw
> 
> ```yaml
> ...
>     steps:
>       - uses: actions/checkout@v6
>       - uses: cachix/install-nix-action@v31
>         with:
>           github_access_token: ${{ secrets.GITHUB_TOKEN }}
>       - run: nix build
>       - run: nix flake check
> ```
> 
> (tags should be digests instead of tags for security reasons but this is the rough shape of what it looks like)



> As a nix flake user of this package I'd love a test that the nix flake compiles, but it's probably too much overhead for the dusklight maintainers if they're not actively using the flake, right?

I'm sure me and the other nix users would pop in to fix it when it breaks - The can message on Discord, or wait for us to pop in when it breaks - Personally, my systems will throw an error every time I update at the moment until this is merged...

@encounter On that note, can you do me a merge please and thank you? (Or PBJ or whoever finds this first?)

**Comment by Krutonium** [comment]
> Another useful check here would be to make sure the flake inputs match the rest of the project, like aurora-src (submodule), dawnlight, imgui, and so on (urls in cmake files, a few in this repo but also a few in the submodule). Or bigger picture figuring out how to get nix-update-script to do this, thinking ahead to getting this in nixpkgs and making sure it's fully r-ryantm compatible.

On the note of nix-update-script, I have *no* idea how to even approach that. Do you know where I might find documentation?

**Comment by voidlily** [comment]
https://github.com/NixOS/nixpkgs/blob/master/pkgs/README.md#automatic-package-updates

I haven't done it either other than the base case of `passthru.updateScript = use-update-script { }`, but you may be able to finagle something together to combine that with "parse the cmake for the up to date urls"? A bit more advanced than anything I've had to do before though so not really in a position to give further advice past that yet.

**Comment by encounter** [comment]
Any way to make it pull from the actual submodule ref in the git repo? I am not going to update this every time I update the aurora submodule.

**Comment by Krutonium** [comment]
As far as I am aware, no, it can't just reference it like that, or it breaks the guarantees of Nix. I think it'd be possible to point nix at the submodule directly; But that'd just eliminated revision/hash updates; It'd still need lockfile updates every time instead.

We're willing to maintain this, all we need is for you to click merge (or give one of us permission to I guess?)

Unless @voidlily has any thoughts on this of course, somthing I've missed?

**Comment by Krutonium** [comment]
Actually :thinking: 
I think if we make a note that you should import it like this (for consumers):
```
dusklight = {
  url = "git+https://github.com/TwilitRealm/dusklight?submodules=1";
};
```
That might solve the issue?

I'm going to have myself a test implementation and get back to you.

Edit: I think this will work, but it's going to dramatically slow down every single update for everyone down stream, as instead of being able to just read the hash, every single computer updating their hashes have to clone the entire repo and calculate it locally, which is... very slow. And Unfriendly for GitHub.

That said, https://github.com/NixOS/nixpkgs/pull/519024 exists; so once it's merged this flake will become less important regardless, except for when people are aiming to run the latest commits.

(Earliest possible time will be the ~25~ 30th, after the the NixOS 26.05 freeze ends)

**Comment by Krutonium** [comment]
For users of the flake:
```
     dusklight = {
       url = "git+https://github.com/TwilitRealm/dusklight?ref=main&submodules=1";
     };
```

**Comment by Mastermindaxe** [comment]
Neither building the pull request directly, nor the main branch with the submodules addition works for me atm :(

Failing with for the `pull/1237`:
```
sed: can't read 76c6lx0ka16gk69b74xy5m32sidcmwzj-source/extern/aurora/CMakeLists.txt: No such file or directory
```

**Comment by voidlily** [comment]
This input worked for me, as the ref `refs/pull/1237/merge` points to "what main would look like after #1237 is merged"

```nix
    dusklight = {
      url = "git+https://github.com/twilitrealm/dusklight?ref=refs/pull/1237/merge&submodules=1";
      inputs.nixpkgs.follows = "nixpkgs";
    };
```

Krutonium:
> This input worked for me, as the ref `refs/pull/1237/merge` points to "what main would look like after #1237 is merged"
> 
> ```nix
>     dusklight = {
>       url = "git+https://github.com/twilitrealm/dusklight?ref=refs/pull/1237/merge&submodules=1";
>       inputs.nixpkgs.follows = "nixpkgs";
>     };
> ```

Comment out the `nixpkgs` follow unless you're on stable or 26.05. Or have it follow your unstable import if you have one.

voidlily:
In my case it's following my unstable import (I do "nixpkgs" is unstable, and "nixpkgs-stable" is stable for whatever reason)

Krutonium:
Amusingly I do the opposite - stable base; with nixpkgs-unstable overlayed onto it under `pkgs.unstable.`

Issue #1228, opened 2026-05-13
[Suggestion] Add cheat to disable the check for the Shadow Crystal


Aepoh:
There's a dolphin cheat for both the Gamecube and Wii versions of the game that lets you transform into a wolf anywhere, anytime. This port sort of already has most features of that, but you still need to get the Master Sword and extract the Shadow Crystal first to actually be able to transform when you want. I would like a cheat that lets you transform before getting the Shadow Crystal.

Justification: It's funny and interesting to turn into wolf link before you're normally allowed to.

SuperDude88:
Added in #1241

PR #1241, opened 2026-05-14
Adds a new cheat that let you transform from the start of the game.

Abzol:
Not fully implemented - Currently does not bypass the check when entering a new zone that sets you to the correct form, unless you edit your Transform Level manually from the save editor.

Allows transformation _only_ using the R+Y quick transform feature, even before meeting Midna or obtaining the Shadow Shard. Intentionally does not alter behavior of the midna menu to avoid possibly breaking any scripted dialogue or interactions before you would normally have the option in the menu.

### DIFF
--- a/include/dusk/settings.h
+++ b/include/dusk/settings.h
@@ -173,6 +173,7 @@ struct UserSettings {
         ConfigVar<bool> fastSpinner;
         ConfigVar<bool> freeMagicArmor;
         ConfigVar<bool> invincibleEnemies;
+        ConfigVar<bool> transformWithoutShadowCrystal;
 
         // Technical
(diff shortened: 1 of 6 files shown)


Abzol:
closes #1228

Abzol:
Of note, enabling this cheat will cause similar "issues" to EMS when it comes to the forced transformations after clearing each twilight level (stay as wolf when supposed to become human) and after finishing lakebed (staying as human when meant to become wolf). That's just how the code is structured.

It will not cause any of the other EMS-related issues (such as breaking sumo), to my knowledge.

encounter (approved):

Issue #1198, opened 2026-05-13
Refactor: split settings.cpp in UI into per-tab methods


wielandino:
src/dusk/ui/settings.cpp is ~1365 lines, with the SettingsWindow constructor alone spanning ~875 lines. All seven tabs (Prelaunch, Video, Input, Audio, Gameplay, Cheats, Interface) are built inline as lambdas inside the constructor body.

This makes the file hard to navigate and review - diffs for a small change in one tab pull the whole constructor into context. 

Proposal: Extract each tab's build logic into a private member function on SettingsWindow (e.g. buildVideoTab, buildAudioTab, …). No behavior change; the constructor just calls them. All code stays in settings.cpp for now. Also some code cleanup in order to fix code duplications

Tell me what you think and I would like to make a PR

PR #1271, opened 2026-05-14
Issue #1198 - split settings.cpp in UI into per-tab methods

wielandino:
Issue: #1198 

Extracts the seven tab build blocks from the `SettingsWindow` constructor
into private member functions. The constructor now only calls them in order.

No behavior change intended
### Testing
- Built with `windows-msvc-debug`.
- Opened the settings window and clicked through all seven tabs.
- Verified a representative toggle in each tab (e.g. Master Volume in Audio,
  VSync in Video, Speedrun Mode in Gameplay) still saves and takes effect.

### Follow-up:
While working through this file I noticed that `settings.cpp` has accumulated
responsibilities beyond just building the UI. A few examples:

- View construction, persistence (`config::Save()`), audio feedback
  (`mDoAud_seStartMenu`), and backend side effects (`aurora_*`) are interleaved
  inside the same lambdas.
- `reset_for_speedrun_mode` is domain logic that happens to live in the UI file.
- The menu-bar rebuild pattern is duplicated across unrelated `onChange`
  callbacks.
- Enum-to-label mappings, help texts, and view bindings sit side by side without
  clear separation.

These feel like real architectural issues rather than just cosmetic cleanup,
and I don't think they belong in this PR. I'd like to open a separate
discussion issue to propose a phased approach - starting with extracting
domain logic (e.g. speedrun mode) out of the UI layer.

Happy to hear if that direction is welcome before I write it up.

### DIFF
--- a/src/dusk/ui/settings.cpp
+++ b/src/dusk/ui/settings.cpp
@@ -20,7 +20,6 @@
 #include "menu_bar.hpp"
 #include "modal.hpp"
 #include "number_button.hpp"
-#include "menu_bar.hpp"
 #include "pane.hpp"
 #include "prelaunch.hpp"
 #include "ui.hpp"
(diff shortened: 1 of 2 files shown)


encounter:
Thanks for the PR, but I don't think this is the direction I'd go to clean this up. I may revisit the settings UI architecture later.

PR #1289, opened 2026-05-14
Set moon jump horizontal speed

RealMochaFox:
Made this primarily for myself but figured I would open a PR in case y'all were interested.

Update the moon jump cheat to also set horizontal velocity and character angle, to provide a bit more control when in the air as opposed to relying on pre-jump momentum for control.


PJB3005 (approved):

PJB3005:
After some internal discussion, I think we've decided against this.

RealMochaFox:
Thanks for considering!

PR #1314, opened 2026-05-14
Update item timer cheat effect 

### LINKED ISSUE #1201: Suggestion:Infinite chu jelly timer
Much like the item timer cheat, apply to chu jelly lying on the ground.

Luzagar:
to fairy tears/ yellow chuchuand hotspring water 
close #1201


MelonSpeedruns:
I think it should be its own cheat named "No Item Timer" and the current setting renamed to "No Drops Timer"

encounter:
After discussion, the team decided that this may be a confusing cheat to add, since Chus and hot spring water are very different types of items

Issue #580, opened 2026-04-28
Moon Jumping underwater


Captainkittyca2:
When moon jumping underwater, you only go up while wearing the iron boots and Link stops going up once he reaches surface of the water.

https://github.com/user-attachments/assets/ede43ef2-382a-4fcb-9ed6-f32010469777

PR #1320, opened 2026-05-14
Fixed being unable to break through the water surface with the moonjump cheat

flash-computer:
Partially addressing #580 
Added the speed vector to debug the former, but I think it's a fair enough metric to be able to see in general since the existing speed variable only tracks the along the surface speed.


flash-computer:
I'm conflicted about cluttering the code with expressions like that, but can't do anything until the input system sees a rework.

flash-computer:
Reverted the debug window changes since #1641 was merged and obsoletes this change.

encounter (approved):

PR #1329, opened 2026-05-14
update ios install instructions

tomlube:
better OS interoperability with iloader and imo much easier to set up too. thanks to seraphie on discord for helping me with this one and looking into a better solution than altstore is


tomlube (commented):
just changing a tutorial

PJB3005 (approved):
I don't know iOS but I trust you enough

PR #1336, opened 2026-05-14
Implemented iPhone haptics rumble option

MiguelBenavides:
A new "iPhone Haptics" setting in Input>Configure controller>Rumble. It only appears if the current device is an iPhone. Useful for controllers that do not have a rumble built-in like the Backbone One. 

If the device is not an iPhone then the option is not compiled into the build.

What changed:

  - Added persisted settings: game.useIphoneRumble and game.iphoneRumbleIntensity in include/dusk/settings.h:157.
  - Added a rumble routing layer in src/dusk/rumble.cpp:18, so controller rumble remains the fallback everywhere.
  - Routed patterned game rumble through that layer from libs/JSystem/src/JUtility/JUTGamePad.cpp:456.
  - Added iPhone-only CoreHaptics support in src/dusk/ios/iphone_haptics.mm:90. It only reports available on actual iPhone devices with haptic hardware.
  - Added the UI under Input > Configure Controller > Rumble in src/dusk/ui/controller_config.cpp:830:
      - Use iPhone Haptics
      - iPhone Rumble Intensity
  - Added iOS-only CoreHaptics build wiring in CMakeLists.txt:556.


encounter:
This should go in aurora, but without AI.

PR #1358, opened 2026-05-15
Soft reset hotkey on release

matthewdavidrichardanderson:
changes the control+r soft reset keybind to activate on release rather than when pressed to match the gamecubes console reset button for speedruns.

from zsr for forest back in time: "Hold the console reset button, and when the continue prompt appears (it has a small delay before it's usable, wait a bit until then) release the console reset button and press A on the same frame."

tested locally and works as intended


encounter (approved):
Thanks!

PR #1372, opened 2026-05-15
fix goron mines elevator path index on Android ARM

### LINKED ISSUE #1105: Crash Goron Elevator
Playing on android base S23.
Before completion of 2nd dungeon and after no matter what I do the game crashes as I take the elevator down.
The issue is the elevator in the entrance to goron mines (The elevator a goron is guarding)
The game runs good otherwise with only one crash as I opened a chest. but this specific loading zone causes a crash each time.

Seems some people on the discord has experienced the same issue but the majority no.

SailorSnoW:
fix https://github.com/TwilitRealm/dusklight/issues/1105

On platforms where plain char is unsigned, the -1 direction value widens to 255 instead of staying −1, so the path index becomes 256 instead of 0 and the chaos happen then.

idk if this is something we are aware of, but that may include other problems in the codebase on ARM devices I think.

Tested on an Android ARM device with the issue savestate.

### DIFF
--- a/src/d/actor/d_a_obj_dmelevator.cpp
+++ b/src/d/actor/d_a_obj_dmelevator.cpp
@@ -596,7 +596,11 @@ void daObjDmElevator_c::moveInit() {
 
 int daObjDmElevator_c::moveProc() {
     dPath* path = dPath_GetRoomPath(getPathID(), fopAcM_GetRoomNo(this));
+#ifdef TARGET_PC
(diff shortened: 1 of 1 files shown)


encounter:
Thanks for pointing this out, we should be building on ARM with -fsigned-char instead of patching it individually though

encounter:
Resolved in https://github.com/TwilitRealm/dusklight/commit/025a425884cb6e9ba8b3f2d85bb465c6973aad66

PR #1386, opened 2026-05-15
Allow Epona to moon jump if the cheat is enabled

flash-computer:
Useful since if epona is moving, her dashes can actually be triggered midair too.


encounter:
I can't get this to work, it seems half the time Link just jumps off of Epona

SuperDude88:
At least for now, we are planning to keep the moonjump cheat the same so that it is consistent with moonjump% category extensions. We appreciate the contribution though -- maybe it could find a place in a mod that expands the cheats?

flash-computer:
That's perfectly reasonable. Thank you for considering this.

PR #1401, opened 2026-05-15
Added a "Turbo Speed" setting to manage the target speed of the game.

### LINKED ISSUE #1276: [Suggestion] Let me choose the multipliers of the turbo function
Up to 4x is not ideal. Put the option to 1.5x, 2x, 2,5x etc
4x is way to fast.

flash-computer:
Value can range from 10% to 2000% at 5% steps.

Additionally, Percentage fields in the settings now automatically snap to the defined step value (rounding around `(step+1)>>1`).

Also made a minor optimization as I stumbled upon it in `src/dusk/game_clock.cpp`. Do tell me if it's out of scope for the pull request and I'll remove it.

Tested and working (For 10% to 400% atleast. My computer can't run it any faster).

Closes #1276

### DIFF
--- a/include/dusk/settings.h
+++ b/include/dusk/settings.h
@@ -179,6 +179,7 @@ struct UserSettings {
 
         // Controls
         ConfigVar<bool> enableTurboKeybind;
+        ConfigVar<float> turboSpeed;
         ConfigVar<bool> enableResetKeybind;
 
         // Tools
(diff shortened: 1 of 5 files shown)


flash-computer:
What is it with these network commands failing so often github?
image


(Don't mind the force push. I'm an idiot who doesn't check what branch he's accessing.)

Issue #1437, opened 2026-05-15
FPS lock


DavCortezzz:
On some devices, the maximum FPS that can be displayed on the monitor is not entirely stable; it would be good to have an option to lock the FPS at 60, 90, 120, 180, etc.
I haven't found a way to lock the FPS on platforms other than PC. For example, on my phone the game runs at 120 FPS, but drops to 100, so locking it at 90 FPS would work for me.

PR #1446, opened 2026-05-15
feat: FPS Limiter

### LINKED ISSUE #747: Add framerate cap option
Please let us set an arbitrary framerate cap instead of being limited to vsync cap

This can be useful on android or laptops to reduce battery consumption and heat.

Also please let us choose between FIFO and mailbox for Vulkan

superyu1337:
Based on work from #1451
Closes #747, #1437 

Implements a frame limiter. Huge thanks to @SailorSnoW who laid the ground work.
This PR moves their limiting logic into the main execution loop and fixes some inconsistencies to improve frametime stability.

# Changes over SailorSnoWs PR
- Moved limiter out of waitForTick. This worked better for Framepacing from my experience.
- Fixed limiter calculating overhead twice with `waitPrecise()`
- More sophisticated `nanoSleep()` which uses a hybrid approach on MacOS.

### DIFF
--- a/include/dusk/settings.h
+++ b/include/dusk/settings.h
@@ -72,6 +72,7 @@ struct UserSettings {
         ConfigVar<bool> lockAspectRatio;
         ConfigVar<bool> enableFpsOverlay;
         ConfigVar<int> fpsOverlayCorner;
+        ConfigVar<int> maxFrameRate;
     } video;
 
     struct {
--- a/include/dusk/settings.h
+++ b/include/dusk/settings.h
@@ -73,6 +73,7 @@ struct UserSettings {
         ConfigVar<bool> enableFpsOverlay;
         ConfigVar<int> fpsOverlayCorner;
         ConfigVar<int> maxFrameRate;
+        ConfigVar<int> maxFrameRatePreset;
     } video;
 
     struct {
--- a/include/dusk/settings.h
+++ b/include/dusk/settings.h
@@ -36,8 +36,8 @@ enum class GyroMode : u8 {
 
 enum class FrameInterpMode : u8 {
     Off = 0,
-    Unlimited = 1,
-    Capped = 2,
+    Capped = 1,
+    Unlimited = 2,
 };
 
 namespace config {
--- a/include/dusk/settings.h
+++ b/include/dusk/settings.h
@@ -34,6 +34,12 @@ enum class GyroMode : u8 {
     Mouse = 1,
 };
 
+enum class FrameInterpMode : u8 {
+    Off = 0,
+    Capped = 1,
+    Unlimited = 2,
+};
+
 namespace config {
 template <>
 struct ConfigEnumRange<BloomMode> {
@@ -58,6 +64,12 @@ struct ConfigEnumRange<GyroMode> {
     static constexpr auto min = GyroMode::Sensor;
     static constexpr auto max = GyroMode::Mouse;
 };
+
+template <>
+struct ConfigEnumRange<FrameInterpMode> {
+    static constexpr auto min = FrameInterpMode::Off;
+    static constexpr auto max = FrameInterpMode::Unlimited;
+};
 }
 
 // Persistent user settings
@@ -73,7 +85,6 @@ struct UserSettings {
         ConfigVar<bool> enableFpsOverlay;
         ConfigVar<int> fpsOverlayCorner;
         ConfigVar<int> maxFrameRate;
-        ConfigVar<int> maxFrameRatePreset;
     } video;
 
     struct {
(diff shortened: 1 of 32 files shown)


encounter:
We already have an FPS limiter in JUTDisplay

superyu1337:
Finally ready for review now.

include/dusk/settings.h
```diff
@@ -123,7 +136,7 @@ struct UserSettings {
         ConfigVar<BloomMode> bloomMode;
         ConfigVar<float> bloomMultiplier;
         ConfigVar<bool> disableWaterRefraction;
-        ConfigVar<bool> enableFrameInterpolation;
+        ConfigVar<FrameInterpMode> enableFrameInterpolation;
```
encounter, line 140: We should make sure this migrates from existing configs properly (true -> Unlimited)

include/dusk/time.h
```diff
@@ -74,7 +81,8 @@ class Limiter {
     if (!initialized || numSleeps++ % 1000 == 0) {
       LARGE_INTEGER freq;
       if (QueryPerformanceFrequency(&freq) == 0) {
-        DuskLog.warn("QueryPerformanceFrequency failed: {}", GetLastError());
+        /// Todo: Make this log work again possibly.
+        //DuskLog.warn("QueryPerformanceFrequency failed: {}", GetLastError());
```
encounter: You can just remove this

include/dusk/time.h
```diff
@@ -99,7 +107,36 @@ class Limiter {
     } while (current.QuadPart - start.QuadPart < ticksToWait);
   }
 #else
-  void NanoSleep(const duration_t duration) { SDL_DelayPrecise(duration); }
+  void NanoSleep(const duration_t duration) { 
+#if defined(__APPLE__)
```
encounter: This would make more sense as an elseif above, as in, define NanoSleep + contents together

libs/JSystem/src/JStudio/JStudio/jstudio-object.cpp
```diff
@@ -655,7 +655,7 @@ void JStudio::TObject_camera::do_paragraph(u32 param_1, void const* param_2, u32
 
 value:
 #if TARGET_PC
-    if (dusk::getSettings().game.enableFrameInterpolation && u <= 5 &&
+    if (dusk::getSettings().game.enableFrameInterpolation.getValue() != dusk::FrameInterpMode::Off && u <= 5 &&
```
encounter: This is feeling like a `dusk::frame_interp::is_enabled` helper would be useful. This is copy/pasted in a _lot_ of places

libs/JSystem/src/JFramework/JFWDisplay.cpp
```diff
@@ -370,28 +370,30 @@ constexpr auto FRAME_PERIOD = std::chrono::duration_cast<std::chrono::nanosecond
 constexpr auto RETRACE_PERIOD = FRAME_PERIOD / 2;
 
 static void waitPrecise(Limiter& limiter, Limiter::duration_t targetNs) {
-    const auto sleepTime = limiter.SleepTime(targetNs);
+    Limiter::duration_t sleepTime = limiter.Sleep(targetNs);
```
encounter: nit: can still use const auto

include/dusk/settings.h
superyu1337: The config parsing step now handles this.
When a cvar is internally changed to an enum, and the parsing step sees a bool coming in, it maps false/true to 0/1 respectively on the enum.

libs/JSystem/src/JStudio/JStudio/jstudio-object.cpp
superyu1337: Changed the code up to use the existing `dusk::frame_interp::is_enabled()` which was never used before for some reason. `dusk::frame_interp::begin_frame` now takes the `FrameInterpMode` enum instead of a bool to set its internal `g_enabled` value.
However this means that there is still once instance of manual comparison left.
We still need to check it the old way within `dusk::game_clock::advance_main_loop()` because otherwise the `dusk::frame_interp::g_enabled` never gets updated.

src/d/actor/d_a_horse.cpp
```diff
@@ -3165,7 +3165,7 @@ void daHorse_c::setReinPosNormalSubstance() {
 #if TARGET_PC
 void daHorse_c::lerpControlPoints(f32 alpha) {
     // FRAME INTERP NOTE: Currently only lerping points for Epona's reins. Need a more global solution.
-    if (!dusk::getSettings().game.enableFrameInterpolation || !s_horseReinSimPrevValid || !s_horseReinSimCurrValid) {
+    if (dusk::getSettings().game.enableFrameInterpolation.getValue() == dusk::FrameInterpMode::Off || !s_horseReinSimPrevValid || !s_horseReinSimCurrValid) {
```
SuperDude88: !dusk::frame_interp::is_enabled() should work here (and for the other similar cases)

SuperDude88 (dismissed):

src/d/actor/d_a_horse.cpp
superyu1337: Done. I think all cases of these manual checks are replaced now, except the one case where it must be used.

encounter (approved):

PR #1447, opened 2026-05-15
Fix item wheel wrong frame with interp

SailorSnoW:
so basically, there was a known problem where the frame on the wheel buffer was wrong when framerate was >30, really confusing for runners coming to dusk as it break cue frame on tricks. In my case, I needed to use the frame before the expected one for the forest bit. 

This aim to fix this using the same presentation-sync method to force the game frame matching console.

Maybe a double-check is required on other tricks/cases.

Here is the trick with the good frame using that fix

https://github.com/user-attachments/assets/aba637db-5f58-4bcf-82ad-4ab84dae7639

### DIFF
--- a/src/d/d_menu_window.cpp
+++ b/src/d/d_menu_window.cpp
@@ -26,6 +26,10 @@
 #include "f_op/f_op_overlap_mng.h"
 #include "m_Do/m_Do_controller_pad.h"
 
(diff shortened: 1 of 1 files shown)


encounter (approved):

encounter:
Thanks!

PR #1448, opened 2026-05-15
Rework No Heart Drops Code

### LINKED ISSUE #1377: Found another room that drops hearts when "Do Not Drop Hearts" is turned on.
V1.1.1. In Hyrule Castle there's a room with a solo Darknut miniboss and a latern oil cauldron. One of the suits of armor next to the cauldron drops hearts when destroyed with the Ball and Chain even though "Do Not Drop Hearts" is turned on

SuperDude88:
- Checks the setting once in `daItem_c::_daItem_create` instead of in every `fopAcM_` helper for creating an item 
- No longer prevents hearts from being created as demo items (should never happen anyways? and would probably break)
- Fixes table entries passed to `fopAcM_createItemFromTable` being incorrectly treated as item IDs
- Prevents any freestanding hearts placed directly in the map with the dzr/dzs from spawning
- Fixes #1377/#1063/should cover more edge cases

Worth doing more testing to validate the behavior changes/make sure no new issues are introduced.

### DIFF
--- a/src/d/actor/d_a_obj_item.cpp
+++ b/src/d/actor/d_a_obj_item.cpp
@@ -268,6 +268,11 @@ int daItem_c::_daItem_create() {
     }
 
     m_itemNo = daItem_prm::getItemNo(this);
+#if TARGET_PC
+    if (dusk::getSettings().game.noHeartDrops && isHeart(m_itemNo)) {
+        return cPhs_ERROR_e;
+    }
+#endif
     BOOL flag = dItem_data::chkFlag(m_itemNo, 2);
 
 #if DEBUG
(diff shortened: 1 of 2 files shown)


SuperDude88:
Known issue with this: hearts for castle town NPCs/end of Snowpeak ruins won't spawn. Not sure if this is fixable with this approach, will look into it.

encounter:
Is this mergable without fixing the known issue?

SuperDude88:
> Is this mergable without fixing the known issue?

I'm think it's fine, as long as the rest of the team doesn't have major concerns

MelonSpeedruns:
I think it's fine to merge it even if hearts don't appear at snowpeak's boss & the fangirls in castle town. Maybe just have a note somewhere as a TODO for those 2?

Abzol:
Does anyone know if they spawn hearts on tphd hero mode? if not then it should probably be considered correct like this

SuperDude88:
> Does anyone know if they spawn hearts on tphd hero mode? if not then it should probably be considered correct like this

From a YouTube longplay it looks like the hearts still spawn from Yeto at least

encounter (approved):

PR #1461, opened 2026-05-16
Update controller bindings menu.

Abzol:
Adds a button that resets the current controller to the default configuration provided by aurora.
Additionally, only renders the panels for Digital L and R if the user has enabled the advanced menu, as there is very little reason a normal user would ever bind these buttons.


encounter:
So going forward, we'd direct people to reset the mappings rather than unbinding the digital triggers? (Have we verified that the reset button does so?)

Abzol:
The reset button does not bind Digital L or R on any controller I tried (and does unbind them), but default bindings come from Aurora.

Abzol:
Though I would like the restore defaults button to play a sound similar to most other menu options, right now there's no indication that you clicked it.

Abzol:
I didnt, in fact, merge from my own PR. I merged from Melon's branch.

PR #1470, opened 2026-05-16
feat: for releases on macos, get directly .dmg

### LINKED ISSUE #1469: Missing .dmg file in Dusklight release
Right now Dusklight in their release misses the .dmg file (only the tar) for the macos machines.
In my opinion this would be better for macOs applications for better integrability.

MagiHotline:
Fixes #1469.


encounter:
The macOS releases are .zip which I prefer over dmg. Additionally, this isn't where we'd package a dmg for releases anyway.

PR #1478, opened 2026-05-16
Infinite item cheats now backup and restore values

ghost:
Infinite item cheats now backup and restore the previous item count.

When cheats were enabled (e.g. arrows or bombs), they'd reset the count to max and you'd lose your previous value.

Now when cheats are enabled:

1. Save the current amount.
2. Keep it maxed while cheat is on.
3. Restore original count when disabled.

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -1,50 +1,164 @@
-### Building
-#### Prerequisites
+# Building Dusklight
+
+## Dependencies
+
+The following dependencies are required:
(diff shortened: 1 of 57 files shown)


SuperDude88:
This seems like it would cause ammo counts to transfer between save files if you enabled the cheat on one file and then disabled it on another. Also would only save your previous count until you restart the program.

PR #1479, opened 2026-05-16
Rename Controller mentions to Device + Extra menu sounds

MelonSpeedruns:
Renames all "Controller" mentions to "Device" to respect Keyboard devices. Also added sounds to the input binding menus.


MelonSpeedruns:
Merge #1461 instead.

PR #1481, opened 2026-05-16
Localization efforts via JSON

sera-phie:
This took me way too long...

This essentially stores all strings into a single JSON file (en_US.json) for people to eventually localize the Dusklight menu. I have not added any menus to switch between languages as there are no other JSON files as of right now. This JSON file can be edited while the game is running and the game will update the strings in real time.

Maddie said we could use the translate bit for RMLUI, but I figured that an RML syntax for localization would be too unintuitive to read & organize, and would grant no inherent benefits, so I stuck with JSON as it's the easiest for others to read and for me to organize.

I have also fixed any additional rebranding holes.

There may have been some places I missed (such as strings from tooltips that I missed in a couple files, or forgetting to update the corresponding C++ file), but 95% of it should be fully usable and complete, and issues should not be immediately visible.


sera-phie:
Just realized I forgot the whole controller config, working on that now.

Abzol:
Suggestions:
The `_()` function, while very small on screen, is hard to know what it does. Something like `LocalizeString()` is probably better.

`pane.add_section(_("device-config.port-d03-h03"));` is not searchable and requires you to cross-reference the localization library to know what it does.
`pane.add_section(LocalizeString("Emulated Trigger Thresholds", "device-config.port-d03-h03"));` is searchable, and more easily understood.

Keys like `device-config.port-d03-h03` are indiscernible at a glance, unless you know that that is "dropdown-03, heading 03", which also can change at any time if new options are introduced, or options are moved around. Something like `config-heading-trigger-threshold` doesnt rely on the layout of the dusklight ui to tell you what it is.

PR #1484, opened 2026-05-16
Add Windows ARM 64 Build

talynone:
With the rising popularity of Windows ARM 64 devices, such as the latest Surface / Snapdragon X laptops, recently released Snapdragon X2 devices and soon to be released Nvidia N1X SOC users are increasingly expecting native applications.

Built and tested the GitHub build Windows ARM artifact from my fork (https://github.com/talynone/dusklight/actions/runs/25963915954) using my Snapdragon X2 Elite Extreme based Asus Zenbook A16 running Windows 11 ARM 26H1. Tested both DirectX12 (default/auto) and Vulkan(glitchy/buggy due to immature Qualcomm Vulkan drivers). 

Notes:

Most of the changes dealt with building dawn for Windows ARM and making sure DirectX12 support was enabled/working.

The Windows ARM artifact built from the dawn-build repo currently produces x64 libs, I have a pull request that fixes that:
https://github.com/encounter/dawn-build/pull/2

denormals.h: Microsoft's MSVC compiler doesn't accept GCC inline assembly syntax, so the file failed to compile for Windows ARM64 builds.

The fix adds an MSVC-specific code path alongside the existing GCC one. It uses Microsoft's _ReadStatusReg / _WriteStatusReg compiler intrinsics (from <intrin.h>) with the ARM64_FPCR register identifier to do the exact same read-modify-write on the same hardware register — just through Microsoft's API instead of inline assembly. Compiler selection is done with #if defined(_MSC_VER) && !defined(__clang__) so clang-cl (which speaks both dialects) continues to use the GCC path. No behavior change on any existing platform.
   
Here's screenshot evidence of it running as a native Windows ARM 64 application:
image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -113,11 +113,90 @@ set_property(GLOBAL PROPERTY PREDEFINED_TARGETS_FOLDER "_cmake")
 if (CMAKE_SYSTEM_NAME STREQUAL Linux)
   set(DAWN_USE_WAYLAND ON CACHE BOOL "Enable support for Wayland surface" FORCE)
 endif ()
+# Force-enable Dawn's D3D backends on Windows. Aurora's vendor branch leaves
+# these to Dawn's defaults, which silently disable D3D12 in some host
+# configurations (observed on Windows ARM64), leaving only the Vulkan
+# backend. Aurora's package/system branches call _aurora_dawn_set_platform_backends()
+# which sets these to ON — mirror that for the vendor path here.
+if (WIN32)
+    set(DAWN_ENABLE_D3D11 ON CACHE BOOL "Enable Dawn D3D11 backend" FORCE)
+    set(DAWN_ENABLE_D3D12 ON CACHE BOOL "Enable Dawn D3D12 backend" FORCE)
+endif ()
+# Additional Dawn vendor-build options that Aurora's branch leaves unset.
+# These mirror a known-working standalone Dawn build config; without them,
+# vendored Dawn on Windows ARM64 ships with a non-functional D3D12 backend.
+set(DAWN_USE_GLFW OFF CACHE BOOL "" FORCE)
+set(TINT_BUILD_IR_BINARY OFF CACHE BOOL "" FORCE)
+if (WIN32)
+    # Ensure Dawn (built via add_subdirectory) honors
+    # CMAKE_MSVC_DEBUG_INFORMATION_FORMAT=Embedded set by our CI preset, so its
+    # debug info format matches the rest of the project.
+    set(CMAKE_POLICY_CMP0141 NEW CACHE STRING "" FORCE)
+endif ()
+# On Windows-on-ARM hosts CMake can leave CMAKE_SYSTEM_PROCESSOR=AMD64 when
+# cmake.exe runs as x64 under emulation, which makes aurora's prebuilt-package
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -113,12 +113,91 @@ set_property(GLOBAL PROPERTY PREDEFINED_TARGETS_FOLDER "_cmake")
 if (CMAKE_SYSTEM_NAME STREQUAL Linux)
   set(DAWN_USE_WAYLAND ON CACHE BOOL "Enable support for Wayland surface" FORCE)
 endif ()
+# Force-enable Dawn's D3D backends on Windows. Aurora's vendor branch leaves
```
encounter, line 116: Submitting these fixes to aurora would be preferred

CMakePresets.json
```diff
@@ -490,7 +490,14 @@
       "inherits": [
         "x-windows-ci",
         "windows-arm64-msvc"
-      ]
+      ],
+      "cacheVariables": {
+        "CMAKE_SYSTEM_PROCESSOR": "ARM64",
+        "CMAKE_TOOLCHAIN_FILE": "",
+        "AURORA_NOD_PROVIDER": "vendor",
+        "AURORA_DAWN_PROVIDER": "vendor",
```
encounter, line 498: We can move these defaults into aurora as well

CMakePresets.json
```diff
@@ -490,7 +490,14 @@
       "inherits": [
         "x-windows-ci",
         "windows-arm64-msvc"
-      ]
+      ],
+      "cacheVariables": {
+        "CMAKE_SYSTEM_PROCESSOR": "ARM64",
+        "CMAKE_TOOLCHAIN_FILE": "",
```
encounter, line 496: ?

talynone: CMAKE_SYSTEM_PROCESSOR: "ARM64" - On the windows-11-arm runner cmake.exe runs as x64 (under emulation), so CMake's auto-detect of the host processor reports AMD64. Aurora's provider scripts use this to build prebuilt-package URLs (dawn-windows-{arch}.tar.gz), and the wrong value meant they fetched x64 tarballs.

CMAKE_TOOLCHAIN_FILE: "" - Solve error because we're building Dawn from source. When dawn is built from source it brings its own copy of Google's Abseil library along with it. When Dawn's CMake tries to create the absl_strings library, vcpkg's wrapper intercepts the call, sees that it already has an absl_strings, and wires the new library to depend on itself, producing the error Target "absl_strings" links to itself and aborting the build before anything compiles. This empties out the vcpkg toolchain inherited from the x-windows-ci parent preset and stops it from interfering with Dawn's bundled libraries. The x86_64 build keeps vcpkg as before since it uses a prebuilt Dawn package that doesn't carry its own Abseil.

talynone:
Hi @encounter If you want me to move some of this stuff to the aurora repo can you deploy a proper arm64 dawn release first? You haven't made one since the arch fix merge (https://github.com/encounter/dawn-build/pull/2).

encounter:
Yes that'd be best. I tried building the dawn repo but it failed. May have to patch it

talynone:
> Yes that'd be best. I tried building the dawn repo but it failed. May have to patch it

Odd when I built it a few days ago it worked: https://github.com/talynone/dawn-build/actions

encounter:
Added in #1778

PR #1491, opened 2026-05-16
Fix mButtonText overflow causing a crash on linux x86_64

### LINKED ISSUE #847: Target Crash in dungeon with game in French
When i target enemies in Goron Mine for more than 1 second it crash to desktop, also when target goron for use shield bash. Try to disable or enable all option but nothing fix this i have to flash targeting for all goron to avoid crash.

i'm on linux and got this error in log:
*** buffer overflow detected ***: terminated

UPDATE
The target crash in all dungeon with the game set on French.

SailorSnoW:
fix #847 

linux x86_64 builds use fortify by default, which make the original strcpy into the 15-byte mButtonText a checked copy. Following the language used (e.g ES/FR have longer string for the shield attack action), the string are >15, causing the buffer overflow.

I think this is better to unify the codebase, and respect the original game code, by just disabling the fortify, as windows/mac/ARM doesnt use fortify by default.

Tested on Ubuntu x86_64

### DIFF
--- a/src/d/d_meter_button.cpp
+++ b/src/d/d_meter_button.cpp
@@ -18,6 +18,9 @@
 #include "d/d_pane_class.h"
 #include "dusk/frame_interpolation.h"
 #include <cstring>
+#if TARGET_PC
+#include "dusk/string.hpp"
+#endif
 
 #if VERSION == VERSION_GCN_JPN
 #define STR_BUF_LEN 528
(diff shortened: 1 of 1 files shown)


Lurs:
You are the goat. 😊👍

PJB3005:
Leaving aside the topic of unifying the codebase, wouldn't it be better for us to fix the problematic string copies as the fix for the bug? We're still having a buffer overflow on that array that should be fixed.

SailorSnoW:
Yes sure, was the first version of the fix I made that I reverted in favor of this one, do we want to keep the fortify flag tho ?

SailorSnoW:
For the fortify flag, I think there's something we need to decide then. Do we want to keep the flag and ignore fortify on the Linux build, at the risk of not surfacing potential buffer overflows from the original game? Or do we keep fortify on, which could mean more potential crashes to fix later, since there might be other hidden buffer overflows in the original game code ?

PJB3005:
I'd definitely be in favor of discovering the bugs. While I guess most of this UB is harmless (because the original game keeps chugging), that doesn't give me any warm comfort on a modern compiler.

The biggest issue, at least for specifically `strcpy`, is that the Sentry crashes we're getting keep not having functional stack traces (as far as I can tell). If something could be done about that I'd be okay with it for sure.

Should be noted, given the amount of times this has happened already, we *do* have a `SafeStringCopy` helper in the code that can be a drop-in replacement when targeting a C array. It'd probably be good to just go through at this point and dump every `strcpy` in the code, where possible. Maybe then it's moot?

SailorSnoW:
Agree with that. I'd also prefer to fix potential UB even if harmless actually. 

Could be a great way to make them safe, or at least crash cleanly with a proper trace instead of silently overflowing. Should probably be its own PR though and not sure how much LoC it'd end up touching, also we'd need to pick the right one between SafeStringCopy and SafeStringCopyTruncate depending on each strcpy's case

PR #1512, opened 2026-05-17
Roll Button/Wolf Transform Override

linkRace:
Adds buttons for "Roll" and "Transform" overrides in Custom Action Bindings.

When set to a button, remove the Roll ability from the A button (text included). All interact actions stay the same (text included), but roll can only be done from the bound button.

This prevents the annoying case of rolling when you wanted to talk to someone/pick up something or vice-versa.

For the wolf/human transform, enables that without needing to talk to Midna first. Note, only works in an area where you can transform, but will no-op instead of having Midna talk to you if you can't transform in that area.


flash-computer:
Great Start to a much needed change.
I think honestly a complete rewrite of the input system, breaking it away from the native gamecube buttons into the underlying verbs to be remapped is in order at this point.

Native controls are always the biggest allure of a port like this, even beyond visual enhancements and performance upgrades.

flash-computer:
The field renames should be contributed to the upstream zeldaret/tp project, I believe.

linkRace:
> The field renames should be contributed to the upstream zeldaret/tp project, I believe.

Submitted here https://github.com/zeldaret/tp/pull/3163

Issue #1088, opened 2026-05-12
Custom Save Game Location


IrukaAkuma:
That would make syncing saves with syncthing much easier, because of that damn android root perrmission block.

lishaoxia1985:
No, I think it can export and import saves is a good suggest.

mturley:
@lishaoxia1985 why not both? Like @IrukaAkuma I would like to sync my save between PC and Android using Syncthing, which requires the save to be in a place accessible by both Dusklight and Syncthing. This is a common pattern. For those who don't care it could remain in app storage by default, and an import/export feature would also be useful. But that would be a different feature than this request.

Issue #1040, opened 2026-05-11
Add the ability to set a folder for storing save data on Android.


Sparcky-Snowtail:
Add the ability to choose a folder for save data on Android, allowing syncing to load files from a different device, drive, or emulator.

Issue #946, opened 2026-05-11
Android Files management


Geov4ni:
Hello. I saw that they updated the Android app and now Dusk appears in the default Android file manager. This has already helped a lot, but unfortunately, because the files are apparently still inside the root folder, other apps can't access the file, or can't write inside the folder.
I tried using Syncthing to sync my save from my PC and found that it can only read and send the file that is on my phone to other devices, but it can't copy files into the root.

One way I thought of doing it temporarily is to sync a different folder, in the Download folder, for example, and then manually copy the save to the correct folder.
(This can be easily done for saves, but for textures or mods it may take longer)

The best thing would perhaps be if the Dusk folder went inside Android/data, as others have already commented.
If that's not possible, maybe an option in the menu to choose a custom folder for the save file?
Congratulations on the excellent work and thank you for your hard work.

Mikelant99:
Yes please let us have access from Android/data or any other folder with easy access, right now I'm having a hard time transfering the texture files to the dusk folder, the transfer just keeps breaking and it is very slow even If I try to transfer every sub folder one at the time, unzipping a file directly also doesn't work because the storage of dusk is so restrictive at the moment

Formedras:
/sdcard/Android/data is not a viable folder for external access. Multiple varieties of Android, including OxygenOS (OnePlus/Oppo), refuse to allow access to that folder. The ideal choice in the current commercial Android environment is to use Storage Spaces to allow the user to select a folder.

sezinlynn:
would be great if Dusk folder is directly under the sdcard or internal storage where file managers have access to. SoH and 2S2H's folders are in that directory and its a lot easier to use Syncthing.

FugueeState:
Agreed, please make the folder either optional or non-root by default so we can properly write to it.

AluminumHaste:
Holy crap, about to go on vacation and want to transfer my savegame from my linux computer to android version and holy shit it's impossible.
Why does this work in SoH just fine?

Formedras:
> Holy crap, about to go on vacation and want to transfer my savegame from my linux computer to android version and holy shit it's impossible.
> Why does this work in SoH just fine?

It shouldn't be that hard. I set the game to save its data in a custom directory on Android (and in Portable Mode on Windows), then I use Syncthing to sync **just the "USA/Slot A" save folder** between my Windows desktop, my Steam Deck, and my OnePlus 12, and all three platforms work fine.
[...]

PR #1520, opened 2026-05-17
Enable Android custom data folder selection

CarlosManuelRodr:
This PR enables the selection of a custom data folder for save files which isn't currently possible as reported on issues #1040
, #1088 and #946 .

Changes included:
- Enabled the Android UI path for changing the data folder
- Added Android MANAGE_EXTERNAL_STORAGE permission
- Added an Android flow to request “All files access” before opening the folder picker
- Added code path to resume the folder selection flow after returning from Android settings
- Improved the Android write-probe error message to point users at the required permission

### DIFF
--- a/platforms/android/app/src/main/AndroidManifest.xml
+++ b/platforms/android/app/src/main/AndroidManifest.xml
@@ -11,6 +11,7 @@
 
     <uses-permission android:name="android.permission.VIBRATE" />
     <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.MANAGE_EXTERNAL_STORAGE" />
 
     <application
         android:allowBackup="true"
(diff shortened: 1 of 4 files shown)


encounter:
This doesn't work as-is because deeper changes are needed to the I/O in Aurora. Raw FS operations don't work against external storage URIs.

CarlosManuelRodr:
This PR is not about passing external storage `content://` URIs into Aurora. What it does is:
1. It uses the Android folder picker to resolve the selected location into a raw filesystem path in `DuskActivity`
2. Then requests a broad MANAGE_EXTERNAL_STORAGE permission
3. After the dialog is closed a raw path is returned

So Aurora still operates on regular filesystem paths. That is why it worked in my testing without additional Aurora changes.
If the desired way to implement this is using a proper SAF/URI-backed Android data folder flow, then yes, this implementation is not ideal and Aurora I/O changes would be needed.

encounter:
Sure, I was trying to avoid the MANAGE_EXTERNAL_STORAGE permission initially, but if it makes things significantly simpler, I'm not strictly opposed

encounter (approved):
I thought about it more and I'm ok with this solution for now. Thanks!

AluminumHaste:
So is this option available in version 1.4.1 or is it coming in a future update?

PR #1527, opened 2026-05-17
Dusklight Online Preliminary test

Tanzalastas:
Early stage test for Dusklight online play. Still needs work and revisions to function properly. Oversight would be much appreciated but will continue to work on peer to peer connect and split screen support in separate branch(?)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -665,3 +665,31 @@ foreach (target IN LISTS BINARY_TARGETS)
         endif ()
     endforeach ()
 endforeach ()
+# cmake/Modules/FindENet.cmake should be added, or manage package parsing directly:
+find_package(Threads REQUIRED)
+
+# Include directories for our custom network source files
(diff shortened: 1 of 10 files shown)


Tanzalastas (commented):
Please let me know if there is more adjustments I should be targeting as this is just preliminary.

SailorSnoW:
don't feel like this should be in the original codebase, if this even exist some day, that should be as a mod imo (+ vibecoded ?)

MelonSpeedruns:
This looks to be vibecoded yes, which we don't want.

Tanzalastas:
> don't feel like this should be in the original codebase, if this even exist some day, that should be as a mod imo (+ vibecoded ?)

Some specific parts yes on vibecode because I didn't know how to execute but mostly no, mostly just "Transform Frame Interpolation" and "NetworkManager::HandlePeerTransform" with tweaks to for improper syntax and had it add comments for me. But yea, I get that it should probably be a separate codebase if mod support is eventually integrated.

PR #1547, opened 2026-05-17
Input Redesign

flash-computer:
Draft PR to overlook input redesign efforts if approved, since I didn't see any other ongoing work for this in other branches..

Plan:
1. Coalesce all inputs under action bindings. This includes the classical controller bindings, which should be preserved as an option with a toggle.
2. Port over the new control options to the RmlUI window under controller configuration. Make the classical controls a separate section there.
3. Track Analog inputs.
4. Track down all flags and button checks in actor code and replace them with these (with conditional macros, ofcourse).

If approved as the right direction to take this in, I'd like to ask @linkRace to bring the changes from #1512 on over to this branch.

### DIFF
--- a/include/dusk/action_bindings.h
+++ b/include/dusk/action_bindings.h
@@ -7,14 +7,104 @@
 namespace dusk {
 
 enum class ActionBinds {
-    FIRST_PERSON_CAMERA,
-    CALL_MIDNA,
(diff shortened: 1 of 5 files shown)


flash-computer:
Suppose not for now then. I'll work on it till it's more mature.

PR #1550, opened 2026-05-17
Fix enemy arrows loading too fast

Pheenoh:
Delays the first enemy arrow creation by 3 frames after each room load to match GC behavior. On GC, the arrow's REL module takes 3 frames to load from disc on the first arrow in a room, while on PC it loads in a single frame. The target angle is set at creation time, so the missing delay makes the first arrow more accurate, causing speedrunners to get hit during the MDH rooftop section when they otherwise wouldn't


PJB3005 (approved):
There's two minor bugs with this code, but I doubt that'll ever be relevant so it's not worth the added complexity to fix:

* If two arrows get spawned at once, the timer would tick faster which wouldn't make sense
* Does the REL get unloaded on console if *reloading* a room (e.g. via voidout)?

PR #1553, opened 2026-05-17
Achievements improvements v2

qwertyquerty:


TakaRikka:
these are prob fine, though did you ever go over the comments for the last big achievements pr? we kinda rushed it for release, but may want to consolidate it now

Abzol:
I was idly talking with Tom about the loach achievement, because a lot of people catch the one in the fishing hole and complain that the achievement didnt pop. I suggested that the best way might be to have a separate achievement for the fishing hole loach, so people can see that there *are* two loaches to catch, not wonder "i caught the loach, why isnt it in my fish journal? bug??"

encounter (approved):

PR #1564, opened 2026-05-17
Merge decomp - UB Fix

PJB3005:


PJB3005:
Oh wait we already have this commit, no need then.

PR #1571, opened 2026-05-18
Fix limits of transform level in savefile editor

Abzol:
Several users have had an issue trying to use the rmlui savefile editor to change their time; touching the transform level at any point after lakebed makes midna permanently disappear. Confirmed that transform level is set to 0xF after lakebed on Dolphin, so it should be settable in Dusklight too.

### DIFF
--- a/src/dusk/imgui/ImGuiSaveEditor.cpp
+++ b/src/dusk/imgui/ImGuiSaveEditor.cpp
@@ -713,7 +713,7 @@ namespace dusk {
                 transformLevel++;
             }
         }
-        if (ImGui::SliderInt("Transform Level", &transformLevel, 0, 3)) {
+        if (ImGui::SliderInt("Transform Level", &transformLevel, 0, 4)) {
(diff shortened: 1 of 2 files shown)


gymnast86:
I think the same change should apply to Twilight Clear Level as well

Abzol:
Checking on a 100% savefile from a console run, Twilight Clear level seems to stay at 3 even at the end of the game. (Including having broken Ganon's barrier.)
Is there a point in the game you know twilight clear level should be 4?

gymnast86:
The game does set it on this line: https://github.com/TwilitRealm/dusklight/blob/main/src/d/d_s_play.cpp#L1191

But looking into it further I'm not sure spawn point 99 in Sacred Grove ever actually gets used. The Master Sword cutscene doesn't trigger it and warping to spawn point 99 just plays a cutscene where nothing happens, so I guess it's probably fine if Twilight Clear Level doesn't go above 3.

PR #1574, opened 2026-05-18
Fix interp jitter at demo00 loop seam

### LINKED ISSUE #1008: Twilight portal visual bug
https://github.com/user-attachments/assets/9f684227-7728-4bde-95fe-69418bc4a42b

SailorSnoW:
fix #1008

### DIFF
--- a/include/d/actor/d_a_demo00.h
+++ b/include/d/actor/d_a_demo00.h
@@ -132,6 +132,11 @@ public:
     /* 0x2900 */ mDoExt_3DlineMat0_c field_0x2900;
     /* 0x291C */ cXyz field_0x291c;
     /* 0x2928 */ u8 field_0x2928[0x292c - 0x2928];
+#if TARGET_PC
+    f32 mDemoFramePrev;
+    bool mDemoFramePrevValid;
+    s8 mDemoSyncTicks;
+#endif
 };
 
 STATIC_ASSERT(sizeof(daDemo00_c) == 0x292c);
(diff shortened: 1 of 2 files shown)


Irastris:
Triggering an entire presentation sync means that on every loop you freeze rendering for potentially several frames, doesn't it? By my count on previous uses, it equates to 4-5 frozen frames when interpolating to 144FPS. I'd argue that's just as distracting as the original problem.

I think I'd prefer a solution be found that doesn't involve that. Another team member suggested "clever Quaternion usage", but I haven't looked into it yet.

SailorSnoW:
I tested it at 120fps, and the freeze is kinda hard to notice, I think you really need to look at it to notice it. So I was considering it visually much better, even tho technically might not be the greatest workaround I agree.


https://github.com/user-attachments/assets/b9a4ce5a-3b8e-4555-8a4d-f34f7e3b7b74

Irastris:
Hard for me to really determine through the low bitrate and framerate of that clip I'm afraid.

I'd definitely like to see a more nuanced solution so I think I'll leave it up for a different team member to decide whether they want to merge this in the meantime or anything.

TakaRikka:
if it doesn't have other side effects, it's a fine enough solution in the meantime imo. maybe just add a TODO that a better solution should be looked into if possible

encounter:
Agreed, a comment or two would be nice to understand what the code is fixing

Irastris (approved):

Irastris:
Sorry, was about to merge and then really thought about it and realized, this triggers constantly during *any* cutscene. There's nothing to actually isolate it to the MoT where there's an actually visible problem, so we'd be introducing tons of pacing inconsistency across the entire game. We definitely need to look for a proper solution without using this one in the interim.

PR #1592, opened 2026-05-18
Make bloom resolution independent and change blending method

s-ilent:
Hi! I've been bothered by the default bloom so I made some changes. I understand some of these changes are subjective. However, I think the main issue I'm trying to address is objectively good. 

Currently, Dusklight's bloom looks different as the user increases the internal resolution, losing much of its range and becoming visually smaller as the internal resolution increases. After some testing, I also found that the blurring had some minor issues with shimmering, which would become visible in areas where there were bright bloomy objects. 

With these patches, bloom is mostly consistent between different resolutions. It looks closer to the vanilla bloom without any shimmering.

| 1x Res | 12x Res|
| --- | --- |
| Screenshot_20260518_224903 | Screenshot_20260518_224900 |

Key changes:
1. Changes how the blur texcoords were calculated so they would take the texture size into account. Previously, they weren't, so the blur wasn't propagating to lower levels of the pyramid correctly. As the pyramid is now blurred as it reaches the lower levels, the upscaling no longer has visible aliasing artifacts. 
2. Locked the size of the per-pass blur. Since this is a blur pyramid, scaling the sample offset will lead to artifacts from oversampling as the underlying pixels don't fit the sample region. Instead, we reinterpret the blur size as a decay factor that adjusts the contribution of each later (wider) blur pass. The overall intensity is then scaled by the other bloom parameters to make it look close to the classic bloom. 
4. The maximum number of blur passes has been increased to 14, and the number of blur passes used scales based on the render resolution. This might be bad for performance, and maybe the initial buffer size should be clamped. But that requires better resampling on the input, which is difficult to write in a fixed function environment. 

These changes are pretty flexible, so I'm hoping they can be adopted even if they need some tweaking to satisfy everyone. Please let me know your thoughts!

| New | Classic |
| --- | --- |
|Screenshot_20260518_233647 | Screenshot_20260518_233645 |
| Screenshot_20260518_234220 | Screenshot_20260518_234222 |
| Screenshot_20260518_234250 | Screenshot_20260518_234252 |
| Screenshot_20260518_234333 | Screenshot_20260518_234335 | 
| Screenshot_20260518_234450 | Screenshot_20260518_234452 |

### DIFF
--- a/src/m_Do/m_Do_graphic.cpp
+++ b/src/m_Do/m_Do_graphic.cpp
@@ -1535,7 +1535,11 @@ void mDoGph_gInf_c::bloom_c::draw2() {
 
         // Successively downsample and apply blurs.
         static int divStart = 2;
-        static int divNum = 6; // inclusive
+
+        int divNum = 6; // inclusive
+        while (divNum < MaxDivNum - 1 && divRects[divNum].h > 8) {
+            divNum++;
+        }
 
         // The original mBlureRatio is multiplied into each sample, of which there are 8 samples originally.
         // This is applied over two passes, the second one with an alpha of 25%; however, the clipping that this introduces is a bit integral to the look,
--- a/src/m_Do/m_Do_graphic.cpp
+++ b/src/m_Do/m_Do_graphic.cpp
@@ -1511,18 +1511,6 @@ void mDoGph_gInf_c::bloom_c::draw2() {
 
         // Setup blur filter TEV.
         GXSetNumTexGens(8);
-
-        u32 texMtxID = GX_TEXMTX0;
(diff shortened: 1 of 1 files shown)


magcius:
Please post a comparison of existing Dusklight bloom to your proposed bloom. Hyrule Castle Town East (in Twilight), Castle Rooftops (in Twilight), and Snowpeak Ruins room 0 are all good examples.

SuperDude88:
Closing for lack of comparisons to Dusklight bloom/can't tell if we want this without having those -- feel free to update with comparisons but currently not planning to change bloom

PR #1599, opened 2026-05-18
Preview texture replacements without the menu obstructing the view 

GiorgioMendieta:
This changes texture replacement toggle from config_bool_select to graphics_tuner_control.
Allows users to view the texture pack being used without the menu hiding it.

| ON | OFF |
| --- | --- |
| Screenshot 2026-05-18 at 18 22 10 |  Screenshot 2026-05-18 at 18 21 59 |

### DIFF
--- a/src/dusk/ui/graphics_tuner.cpp
+++ b/src/dusk/ui/graphics_tuner.cpp
@@ -4,6 +4,7 @@
 #include "m_Do/m_Do_audio.h"
 
 #include <aurora/aurora.h>
+#include <aurora/gfx.h>
 #include <dolphin/gx/GXAurora.h>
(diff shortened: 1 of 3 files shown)


Irastris (dismissed):
The problem with backing this by an enum is that it doesn't account for partial texture packs, such as those that only impact the HUD, or a single character.

We have a modding API in the works that would facilitate loading texture packs, at which point textures would become a simple matter of installation, toggling, and prioritization from a separate mod loader menu.

Making this option into a GraphicsTuner is still alright, but I'd like if you dropped the enum backing and just treated it as a simple On/Off boolean. After that, this should be mergeable 👍

encounter (approved):

PR #1601, opened 2026-05-18
Rework nix flake for offline multi-platform builds

SailorSnoW:
It add missing platforms for the builds and fix the existing one.

tested build on darwin-aarch64, the other ones evaluates with success so unlikely to not build ig

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1775710090,
-        "narHash": "sha256-ar3rofg+awPB8QXDaFJhJ2jJhu+KqN/PRCXeyuXR76E=",
+        "lastModified": 1778869304,
+        "narHash": "sha256-30sZNZoA1cqF5JNO9fVX+wgiQYjB7HJqqJ4ztCDeBZE=",
         "owner": "nixos",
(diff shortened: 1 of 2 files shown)


encounter (approved):

Krutonium:
This shouldn't have been merged as is - I only just saw it when I went to do a build; but basically as was seen [here](https://github.com/TwilitRealm/dusklight/pull/1237#issuecomment-4457087018) - Aurora was intentionally not part of the build; you need to add it to your flake using submodules as part of the URL, as the developers don't want to merge an update to the flake every time they update Aurora.

Also was it intentional for the resulting derivation to be FAR larger than it used to be? The Binary is almost 500MB instead of the previous ~50MB. (IIRC that's caused by packing in a bunch of debugging information)

Before:
```
✔ >>  ls -l
.r-xr-xr-x 52M root 31 Dec  1969 󰡯 dusklight
dr-xr-xr-x   - root 31 Dec  1969  res
```


After:
```
store/30iicm9ly3hhpxmq948c5dmbg779xyyg-dusklight-nix-1711251/bin🔒 
✔ >>  ls -l
.r-xr-xr-x 454M root 31 Dec  1969 󰡯 dusklight
dr-xr-xr-x    - root 31 Dec  1969  res
```


Unless @encounter has changed their mind?

PR #1617, opened 2026-05-18
three heart challenge fix

tomlube:
checking for 19 health gives you 4 heart pieces and 3 heart containers, which does not affect gameplay


qwertyquerty:
already done in #1553

PR #1634, opened 2026-05-19
Windows-only Autoupdater

superyu1337:
Draft PR for the windows-only autoupdater.


SuperDude88:
Closing since this is still draft/seems mostly stagnant + probably wouldn't want an updater that only works on one platform (though exactly which platforms one should cover is still kinda in the air)

PR #1648, opened 2026-05-19
Input Rework

flash-computer:
Reopening this Draft PR to work on the input redesign.

Some work has been done, but a lot remains. Currently, the button verbs have been reworked into Link's combat and general movement code, but there's still issues like the game tracking actions by buttons and use flags which the rendering unit uses to dim out item icons, which I'm working on fixing next.

A lot of testing would be needed for this, so if you can build the game and test it out (I have added a few options, including input fallthrough from the menu and toggles for both classic and new button inputs in the menu), that'd be helpful later, but for now the verbs have only been really separated, and button statuses hinder proper use.

Hopefully this slightly more mature PR is allowed to stay on and be worked on.

Takes a different route (Jump and Roll actions are under the same verb, pickup, grab and interact are their own different verbs) from #1512, and thus obsoletes it if approved. Have merged the early changes from @linkRace  though.

I'll remove the excessive comments and some Macros in the final PR.


humzakh:
I'm curious why this PR was closed? It would be extremely useful to bind buttons to certain actions vs. what the game requires. Example, changing the sword attack binding from B to X (Xbox/Steam Deck controls) currently makes going back in menus also require using X instead of B. This is unintuitive, unlike the BotW control scheme.

flash-computer:
> I'm curious why this PR was closed? It would be extremely useful to bind buttons to certain actions vs. what the game requires. Example, changing the sword attack binding from B to X (Xbox/Steam Deck controls) currently makes going back in menus also require using X instead of B. This is unintuitive, unlike the BotW control scheme.

It's closed because for now, my ability to work on it is extremely limited by circumstances irl. I will pick it up later and continue working on it, but it might be a while before that.

PR #1654, opened 2026-05-19
Add equipment deselection toggle for swords, shields, and clothes

xiedada05:
Allows deselecting equipped swords, shields, and clothing from the Collections menu when enabled in settings. This reverts Link to bare hands or Ordon Clothes respectively. Fixes the auto-switch to Kokiri in d_a_alink.cpp when the feature is enabled, as Nintendo assumes Link will never wear Ordon Clothes after getting his green tunic, but conflicts with this function.

AI is only used in code navigation, Git commit message and the previous paragraph of this PR. No AI generated code is included in this commit.


MelonSpeedruns:
I coded a very similar mod during Dusklight's development at #361 but ended up canning it due to it feeling more like a mod. Maybe we've changed our minds since then? Idk.

Also, I think if instead of 3 separate options, you had 1 option with 3 toggles inside (Like Gyro mode) we could see about adding it.

Also there are a few crashes when having the Ordon Clothes equipped, so maybe look into fixing those as well! ^^

xiedada05:
Could you please provide the step to reproduce or stack trace? I will try it tomorrow. I did not experience any crashes when testing on Windows 11 x64 w/ D3D12, compiling with Visual C++.

superyu1337:
I'm currently in the middle of writing a model override system that changes the actual model. That would work fine to play through with the entire game because by simply setting the active clothing like you do ATM, you "break" the checks that the game has in place. You'll also have an issue when you need Zora armor for example.
As extended base game feature this is nice to have, but most people want to play through the game with one model equipped.

xiedada05:
So I used menu option to control it, and if model replacement feature is merged to main, we could disable this menu option to prevent conflict.

superyu1337:
Technically there shouldn't be a conflict. The model override works on all outfits in the game, it can even override link to wear a Zora outfit in the sumo fights.
I believe its more of an issue whether the model override gets merged, or if this gets merged. But that's for the Dusklight team to decide.

Edit:
MelonSpeedruns and I just had a little talk about it. Maybe it's better if this PR gets merged instead of my model overrider. I could refactor my code to eventually move it to an official modding feature to load custom models etc.

xiedada05:
@MelonSpeedruns I updated menu to make it cleaner, and polished the logic for handling clothes: when Link has Hero's Clothes, wearing Ordon clothes and disabled deselest clothes, he will instantly put on Hero's Clothes once Collections menu is opened or the scene transition is occured. If there is still crash, please tell me steps to reproduce or give me stack trace.

xiedada05:
The development plan in #1658 has been influenced by #1654 while this PR seems to be forgotten, so I earnestly request the maintainers to review this change and/or implore a prompt response. @SuperDude88

src/d/d_menu_collect.cpp
```diff
@@ -1208,28 +1231,62 @@ void dMenu_Collect2D_c::changeSword() {
...
+            setEquipItemFrameColorSword(0);
+            Z2GetAudioMgr()->seStart(
+                Z2SE_SY_ITEM_SET_X, NULL, 0, 0, 1.0f, 1.0f, -1.0f, -1.0f, 0);
+            dMeter2Info_set2DVibration();
+        }
+#endif
         break;
     case 4:
-        if (dComIfGs_isItemFirstBit(dItemNo_LIGHT_SWORD_e)) {
+        if (dComIfGs_isItemFirstBit(dItemNo_LIGHT_SWORD_e)) {
```
xiedada05:
```suggestion
        if (dComIfGs_isItemFirstBit(dItemNo_LIGHT_SWORD_e)) {
```
Delete uncecessary whitespaces created by Visual Studio

SuperDude88:
This is something we think would make more sense as a mod using the new API

PR #1691, opened 2026-05-20
More Magic Armor Options

### LINKED ISSUE #1196: Suggestion: Magic Armor option
Instead of no rupee drain option, make it to where the over time rupee drain is disabled, but when you take damage rupees are lost.

### LINKED ISSUE #1687: [Enhancement] Magic Armor drains Rupees on hit instead of over time
Is your feature request related to a problem? Please describe.
The vanilla passive rupee drain makes the Magic Armor annoying to use. However, the current "Free Magic Armor" cheat makes it completely overpowered by removing the cost entirely.

A simpler, balanced alternative: change the trigger so the armor only drains rupees when Link actually takes damage.

Remove the passive drain over time.

Subtract rupees only upon taking a hit (scaling with the damage taken or just a flat amount).

Let the vanilla engine handle the rest. (If rupees reach 0, the game naturally takes care of making the armor heavy and removing the invincibility).

Additional context
This just shifts the rupee cost from a timer to an on-hit event. It keeps the game balanced, makes the armor viable for combat, and maintains the vanilla penalty for running out of money.

SuperDude88:
Adds new options to control the magic armor behavior. New choices are:
- original/unmodified
- only lose rupees on damage
- double defense (does not use rupees)
- always invincible
- cosmetic

Closes #1196

### DIFF
--- a/include/dusk/settings.h
+++ b/include/dusk/settings.h
@@ -37,7 +37,9 @@ enum class GyroMode : u8 {
 enum class MagicArmorMode : u8 {
     NORMAL = 0,
     ON_DAMAGE = 1,
-    NEVER = 2,
+    DOUBLE_DEFENSE = 2,
+    INVINCIBLE = 3,
+    COSMETIC = 4,
 };
 
 namespace config {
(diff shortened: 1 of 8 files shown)


MelonSpeedruns:
Closes #1687

encounter:
What needs to be done to bring this out of draft?

encounter (approved):

PR #1692, opened 2026-05-20
maybe fix ook crash

### LINKED ISSUE #1774: Game crashing at multiple different spots in the room with the monkey bridge to ook
As the title says the game is crashing at multiple different spots in the room with the monkey bridge to ook whether its just entering the room or voiding out in the room, it keeps crashing there during my runs.

qwertyquerty:


qwertyquerty:
Might need @Pheenoh or @spicyjuice04 to try this since I can't repro

Pheenoh:
> Might need @Pheenoh or @spicyjuice04 to try this since I can't repro

I can't repro either starting from a forest bit save. Might need to be played from the start of the run to repro

Pheenoh:
> > Might need @Pheenoh or @spicyjuice04 to try this since I can't repro
> 
> I can't repro either starting from a forest bit save. Might need to be played from the start of the run to repro

I'll do a run up to GB this weekend while running gdb if nobody reproduces it by then

spicyjuice04:
Not necessarily need to be from the start @Pheenoh it happened to TheEtherNetBoyz two times in a row in this stream https://www.twitch.tv/videos/2774681116 though it needs to also be considered that they used a build with the insta load which might have had influence, but apparently it is not an influence to others who used it

matthewdavidrichardanderson:
i was testing this fix and got the same type of crash but instead of on the cutscene skip it happened when i voided out to lower boss flags to try again, on 1.2 i also had it crash when i opened the door to go back inside so im assuming all these crashes are related. video starts after i got boss flags

https://github.com/user-attachments/assets/4b4cfe5a-581f-4d2f-ba6b-cf3219ea7031

SuperDude88:
Superseded by #2034?

Irastris:
I think so, gonna close for now

PR #1699, opened 2026-05-21
Modified the Bigger Wallets setting to add more Wallet Sizes options

DarkZant:
Added a Quality of Life setting that disregards wallet limits and allows the rupee count to go to 9 999. This can allow players to collect rupees without having to worry about their wallets being full. A nice quality of life setting for late game with the Magic Armor or simply to try to collect all the available rupees in Hyrule.


DarkZant:
I modified my original setting to make a much cleaner setting which includes the original bigger wallets setting and other options.

DarkZant:
### Wallet sizes options are now:
- Default: 300, 600, 1000
- HD: 500, 1000, 2000
- Large: 1000, 5000, 9999
- Uncapped: Always 9999, no matter the wallet

PR #1704, opened 2026-05-21
Fix mirror mode sign arrows

### LINKED ISSUE #1182: Sign arrows not flipped mirror mode.
When in Faron woods, you come up to a sign that has two arrows. One pointing left for Forest Temple Path, and another pointing right for the pitch dark cave; the arrows do not flip when switching over to mirror mode. I imagine this is the case for similar signs in the game.

itssecretnate:
Doing some more testing to make sure this doesn't affect anything else, but I'm just going to follow the Wii U's behavior for handling this which is just flipping the arrow that are rendered.

image

image

Dusklight:
image

image

resolves #1182


itssecretnate:
Investigating different signs in the game Lake Hylia so far has one that this won't work for.

Wii U with mirror mode:
image

itssecretnate:
Sumo tutorial in Bo's basement.
image

itssecretnate:
I made an assumption while playing TPHD earlier and obviously was wrong. Going to do some more research on this and see if there's possibly another solution.

Like was mentioned in the original issue; if GameCube has the Wii version's mirrored text, I'd like to see if there's a way to use that in a way that won't negatively affect item control descriptions.

jRibe115:
Sometimes it's also Midna telling you to go East or West

itssecretnate:
> Sometimes it's also Midna telling you to go East or West

Do you have an example of this I can add to a list? I'm currently compiling a list of dialogues that are different with the idea being to load while mirror mode is enabled.

jRibe115:
No, unfortunately. Sorry

itssecretnate:
Seeing soft locks in certain dialogue prompts. I have an idea on how to fix it just need to do some testing.

TakaRikka (approved):

Issue #1703, opened 2026-05-21
(Feature) Ability to disable/customize dualshock 4 led light


CloudyCrayons:
It's cool that you have native dualshock 4 support but I would like the ability to dim or turn off the light.

the light notoriously uses up more battery life than you'd expect

SonicRay100:
Was going to suggest this myself. But I was going to request a full LED customisation panel, where you can turn off, dim and even alter the colour for various situations as well.

SonicRay100:
Seems like this is being worked on #1705. But it's only enable/disable. It's a start lol.

SuperDude88:
Option to disable was added with #1705. Feel free to open an issue specifically for customization, though that might be something better handled by a mod once that API is available.

SonicRay100:
Unfortunately, this doesn't turn off the LED when the option is disabled. It won't change anymore but it stays at whatever colour it was when it turned on.

Also, even if the option is disabled, the LED will still turn on when Dusklight is opened.

superyu1337:
The feature added is meant to disable dusklight's own override.
If you disable it, you should be able to disable it change the light yourself via steam for example.

SonicRay100:
I get it, but the fact that it enables the LED when the option is disabled is a bit annoying.

superyu1337:
> I get it, but the fact that it enables the LED when the option is disabled is a bit annoying.

That indeed shouldn't happen. I'll look into it when I have time.

SonicRay100:
I should clarify that I meant when Dusklight is opened with the LED option disabled, the LED activates.

PR #1705, opened 2026-05-21
feat: Make gamepad status lighting optional

superyu1337:
This creates an option for enabling/disabling the gamepad status lighting. The default is true, to not change behavior. But it allows a user to disable the option, should they wish to change the gamepad's lighting or turn it off entirely.


MelonSpeedruns:
Maybe rename it to "Enable Controller LED" ? Sounds easier to understand

superyu1337:
> Maybe rename it to "Enable Controller LED" ? Sounds easier to understand

Done

MelonSpeedruns (approved):
LGTM!

Abzol:
Does this option only show when a LED is detected?

superyu1337:
> Does this option only show when a LED is detected?

No, I could add that tho.

MelonSpeedruns:
> > Does this option only show when a LED is detected?
> 
> No, I could add that tho.

Make sure it shows up and not when plugging in/out the controller.

superyu1337:
converted to draft until relevant [PR](https://github.com/encounter/aurora/pull/205) in Aurora is merged and dusklight is updated.

src/dusk/ui/bool_button.cpp
```diff
@@ -29,7 +30,10 @@ bool BoolButton::disabled() const {
 }
 
 Rml::String BoolButton::format_value() {
-    return mGetValue() ? "On" : "Off";
+    if (this->mIsSupported())
```
encounter: I'd rather just disable the button rather than changing the label, or remove the  button entirely when it's not supported

src/dusk/gamepad_color.cpp
```diff
@@ -5,102 +5,124 @@
...
+            );
+        }
+    }
+}
 
-    currentGamepadColor = LerpColor(currentGamepadColor, cXyz{finalRed, finalGreen, finalBlue}, lerpSpeed);
-    PADSetColor(PAD_CHAN0, (u8)currentGamepadColor.x, (u8)currentGamepadColor.y, (u8)currentGamepadColor.z);
-}
\ No newline at end of file
+}  // namespace dusk::input::gamepadLed
```
encounter: outdated (can use `clang-format -i src/dusk/gamepad_color.cpp` to fix)

src/dusk/ui/bool_button.cpp
superyu1337: I think it would be weirder to disable the button, as it may show "On" and leaves the user confused why they can't change it. I added the feature to the UI element because I thought it may be useful in the future aswell.

encounter: I'd rather not show the button in that case

superyu1337: Alright, I'll update it later.

SonicRay100:
There's a bit of a problem when it comes to disabling the LED. https://github.com/TwilitRealm/dusklight/issues/1703#issuecomment-4724499348

PR #1714, opened 2026-05-21
Update build.gradle to reflect correct Version Name

benyjr:
versionName should match the actual version number so the Android system can display the current software version without having to actually launch the program to find it.


encounter:
Added in #1593

PR #1718, opened 2026-05-21
linux: Fix Wayland window associations

ReillyBrogan:
`src/m_Do/m_Do_main.cpp` uses `SDL_SetAppMetadata("Dusklight", DUSK_VERSION_STRING, "dev.twilitrealm.dusk")` which results in the Wayland appId getting set to `dev.twilitrealm.dusk`. Wayland compositors associate Wayland windows to desktop files by comparing the window appId against installed desktop files by the name of the desktop files. For a given window with appId `dev.twilitrealm.dusk` it would look for `dev.twilitrealm.dusk.desktop`, and because the desktop file is not named that the compositor is unable to correctly associate the window which results in the default Wayland icon being shown for the window. It also results in windows not being associated with the given launcher if the desktop file is pinned to a task management widget of some kind.

Fix this by renaming the desktop file to `dev.twilitrealm.dusk.desktop`. Also while we're at it we can rename the icons as well to match the desktop file name, which isn't required but is a bit cleaner and will fit various packaging standards better.


CMakeLists.txt
```diff
@@ -303,7 +303,7 @@ include(files.cmake)
 # TODO: version handling for res includes
 
 set(DUSK_BUNDLE_NAME Dusklight)
-set(DUSK_BUNDLE_IDENTIFIER dev.twilitrealm.dusk)
```
encounter, line 320: Please revert the bundle ID change, we need it to remain consistent for Android and iOS in particular

encounter (approved):
Thanks!

PR #1728, opened 2026-05-22
Update settings.cpp with more descriptions 

SealEnthusiast:
Adds more descriptions for some settings.

- Enable Depth of Field: Enables a Depth of Field effect when obtaining an item
- Enable Mini-Map Shadow: Adds a shadow effect to the Mini-Map
- Disable Cutscene Pillarboxing: Removes Pillarboxing in some cutscenes

First time editing a file like this, so If I missed anything please let me know.


dooplecks:
Superseded by 6a52cdf.

PR #1738, opened 2026-05-22
Feat: Added the option to change the save file location

Timtendo12:
Added the option that allows you to change the save directory of the save folder without having to drag the assets and other files in the data folder.

By default it saves to the data folder like how it saves it now. When changing the folder it will migrate the save file to the specified folder on a restart.

Tested on windows. Will also test on macos later.


Irastris:
> Note that we do not accept contributions that are primarily AI-generated and will close your PR if we suspect as much.

PR #1744, opened 2026-05-22
build.yml revisions

Irastris:
Made a few changes to build.yml to hopefully improve the CI experience at least a little:

1. Only process push on the main branch, preventing CI runs from doubling up when new branches are created and pull requests are opened
2. Only process pull_request when they intend to merge into main. The types ensure that CI still reruns when an already opened PR receives new commits
3. Use concurrency to automatically cancel running workflows in a pull request if new commits are made before the old ones have finished

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,10 +2,23 @@ name: Build
 
 on:
   push:
+    branches:
+      - main
     paths-ignore:
       - '*.md'
       - '*LICENSE'
   pull_request:
+    branches:
+      - main
+    types:
+      - opened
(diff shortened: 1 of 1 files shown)


encounter:
For 1, it's probably better overall to just fork the repository and submit PRs rather than pushing to the main repo and disabling non-main branch builds. I use branches for testing CI changes and such, and want builds to run.

3 is a good idea tho

Irastris:
Fair enough, I'll revert so only the concurrency fix remains in a couple minutes

encounter (approved):

PR #1746, opened 2026-05-22
feat: Allow line breaks in UI panes

GiorgioMendieta:
Allow line breaks in UI panes, might be useful like in the following example:

Screenshot 2026-05-22 at 20 25 03


SuperDude88:
Text can already have line breaks using `<br/>`

PR #1761, opened 2026-05-23
Change default log level from DEBUG to  INFO

### LINKED ISSUE #1000: [Request] Option to limit log verbosity or disabling log creation entirely
It seems the default logging level is set to `DEBUG` which means that every thing gets logged, including constant Discord rich presence calls.

While I appreciate the reason for it, it would be nice to be able to either limit logs to just warning/error/critial, or disabling them entirely. Normally I would just symlink the logs to `/dev/null` but since a new log is created for each launch with a different timestamp that's not really possible.

The option could be put under advanced settings or just be made changeable by editing the config file.

GiorgioMendieta:
Changes default log level from 0=LOG_DEBUG to 1=LOG_INFO, preventing cluttering the journal.

I think this is useful for end users, during development devs can just open the app with the `-l 0` flag anyways.

Closes https://github.com/TwilitRealm/dusklight/issues/1000

### DIFF
--- a/src/m_Do/m_Do_main.cpp
+++ b/src/m_Do/m_Do_main.cpp
@@ -489,7 +489,7 @@ int game_main(int argc, char* argv[]) {
         cxxopts::Options arg_options("Dusklight", "PC Port of a classic adventure game");
 
         arg_options.add_options()
-            ("l,log-level", "Log level from " + std::to_string(AuroraLogLevel::LOG_DEBUG) + " to " + std::to_string(AuroraLogLevel::LOG_FATAL), cxxopts::value<uint8_t>()->default_value("0"))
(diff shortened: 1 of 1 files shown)


JoonasPitk:
Does this still write at debug level into the .log files? Those would still be useful, but being able to limit system journal spam is definitely needed.

GiorgioMendieta:
> Does this still write at debug level into the .log files? Those would still be useful, but being able to limit system journal spam is definitely needed.

No, log files are written at INFO level by default with this change, unless you specify the DEBUG log level with the flag`-l 0` when launching the game.

I still think it is a sensible default for most users. If needed the log level can be changed to DEBUG (or other levels) for development or debugging. This prevents creating huge log files as well depending on the playtime.

PR #1767, opened 2026-05-23
macOS: Open data folder from menu bar

GiorgioMendieta:
Allows macOS builds to open the Data folder directly from the menu bar instead of going through the in-game options.

Users can click on Dusklight > Open Data Folder... as an alternative of the in-game menu option.

https://github.com/user-attachments/assets/2e15b6f3-baed-4f0f-b99b-2a87774f9ed8


j4bush14 (approved):
This looks good. I think this will be a good option to have for Mac users.

PR #1784, opened 2026-05-24
Allow Sun's Song & Quick Transform to be bound via the Custom Action …

sera-phie:
…Bindings menu along with default.

This is for people that get irritated with not being able to hit the buttons at the exact same time, without also requiring that Midna be rebound.

EDIT: I guess that this isn't needed like at all since it's not a timing thing at all and it was digital triggers messing things up for me, so this basically doesn't need to be done at all. Can use it if you want to though as a general convenience thing.

EDIT: Added moon jump as well, and disabled the default binds when new ones in Custom Action Bindings are bound

### DIFF
--- a/include/dusk/action_bindings.h
+++ b/include/dusk/action_bindings.h
@@ -13,6 +13,7 @@ enum class ActionBinds {
     TURBO_SPEED_BUTTON,
     QUICK_TRANSFORM,
     SUN_SONG,
+    MOON_JUMP,
     COUNT,
 };
 
--- a/include/dusk/action_bindings.h
+++ b/include/dusk/action_bindings.h
@@ -11,6 +11,8 @@ enum class ActionBinds {
     CALL_MIDNA,
     OPEN_DUSKLIGHT_MENU,
     TURBO_SPEED_BUTTON,
+    QUICK_TRANSFORM,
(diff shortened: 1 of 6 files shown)


Abzol:
Rather than
```
if (mDoCPd_c::getHoldR(PAD_1) && mDoCPd_c::getTrigX(PAD_1) ||
    IF_DUSK(dusk::getActionBindTrig(dusk::ActionBinds::SUN_SONG, 0))) {
```
Maybe something like
```
if (dusk::getActionBindTrig(dusk::ActionBinds::SUN_SONG, 0) ||
   (dusk::getActionBindButton(dusk::ActionBinds::SUN_SONG, PAD_NATIVE_BUTTON_INVALID) &&
    mDoCPd_c::getHoldR(PAD_1) && mDoCPd_c::getTrigX(PAD_1))
```

(I havent actually tested if this works)

Abzol:
my bad you want to use `dusk::isActionBound(dusk::ActionBinds::<action bind>, 0`, not compare to button_invalid

MelonSpeedruns:
I think this is neat, but it should probably wait until we put in place a system where action bindings can be mapped to more than 1 button to create a button combo.

PR #1791, opened 2026-05-24
store config to disk when destroying ui

Abzol:
certain ui actions do not call config::Save(), e.g. unbinding action keys. rather than forgetting anywhere else, just save when closing the app


encounter:
I also added a save when hiding the controller window in https://github.com/TwilitRealm/dusklight/commit/2e1cc7cb26fcbac5ae99e41567064a7a395a41ee but having it again on shutdown is probably good too

encounter (approved):

PR #1826, opened 2026-05-26
Restore Green Chu Merging

ZipoLabs:
Restores Wii/HD Green Chu merging behaviour, possible in the Cave of Ordeals. A setting has been added to the "General" section of the "Gameplay" tab.

I ultimately opted not to roll this into the existing "Restore Wii 1.0 Glitches" setting, as Green Chus' presence in the Wii version seems to be more of a remnant of the cut magic meter feature rather than being an outright "glitch".

### DIFF
--- a/include/dusk/settings.h
+++ b/include/dusk/settings.h
@@ -230,7 +230,6 @@ struct UserSettings {
 
         // Technical
         ConfigVar<bool> restoreWiiGlitches;
-        ConfigVar<bool> restoreGreenChuMerging;
 
         // Controls
         ConfigVar<bool> enableTurboKeybind;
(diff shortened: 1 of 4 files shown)


encounter:
Let's bundle this under "Restore Wii 1.0 Glitches", I think it makes most sense there. Otherwise LGTM

mliquid19-cmd:
> Let's bundle this under "Restore Wii 1.0 Glitches", I think it makes most sense there. Otherwise LGTM

Hi Devs. There is a rendering bug when you go to Ordon Village using a horse. My phone gets black screen. When I walk, the black screen doesn't appear. It's the same at Ordon Farm, after I guide the sheep to the barn, when I go out the gate with my horse, the black screen appears again and I have to walk again. It's just annoying. Please fix it.

PR #1827, opened 2026-05-26
[Randomizer] No logic, Random spawn point

JaxonKEKW:
- Implements No Logic (Just 2 changes to fill.cpp, idk if I'm just stupid but it didn't seem to be working at all before, wasn't checked anywhere) _*Spheres still get generated, however it's clear that no logic is now working_

- Fills in a few entries in entrance_shuffle_data and fixed typos (?)

- Implements Random Spawn Location (Gets warp data from yaml and warp menu data, validates and sets next stage)

- Tested a loop to validate all spawns in the pool and only 6 of them fail, which are the ones with invalid data in the yaml (Wii 1.0 exclusive, Fado's house, one-way entrances, etc) and they all get skipped as spawn candidates. Filled in the few that exist

I'm unsure how to structure rando PRs because it's all still super in progress, and I'm sure others are working on both of these. I've been running seeds with this code for a couple weeks and it's been solid. Random spawn currently skips the prologue cutscene however I also tested keeping the cutscene and it would work fine

### DIFF
--- a/include/dusk/io.hpp
+++ b/include/dusk/io.hpp
@@ -31,6 +31,11 @@ public:
     ~FileStream();
 
     /**
+     * \brief Flush buffered writes and throw if the flush fails.
+     */
+    void Flush();
+
+    /**
      * \brief Open a file for reading at the given path.
      */
     static FileStream OpenRead(const char* utf8Path);
--- a/src/d/d_com_inf_game.cpp
+++ b/src/d/d_com_inf_game.cpp
@@ -33,6 +33,7 @@
 #include "dusk/randomizer/game/stages.h"
 #include "dusk/randomizer/game/flags.h"
 #include "dusk/randomizer/game/randomizer_context.hpp"
+#include "dusk/randomizer/generator/randomizer.hpp"
 #endif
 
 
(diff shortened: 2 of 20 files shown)


gymnast86 (changes requested):
While this PR looks perfectly functional, I had some different ideas in mind for the in game entrance rando implementation. 

Right now this PR hardcodes overriding the starting spawn. I'd instead prefer a generic system for overriding entrances in game by creating something like this in the `RandomizerContext`:
```cpp
std::unordered_map<u32, StartLocation> mEntranceOverrides{};
```
where the u32 key is potentially:
```cpp
u32 key = 0;
key |= stageId << 16;
key |= room << 8;
key |= spawn;
```
Then we just override the entrance data somewhere like `dStage_startStage_c::set` if the randomizer is active and an override exists.


The data fields in `entrance_shuffle_data.yaml` that are currently strings don't actually need to be strings. That was just an product of me putting empty strings in for the fields which didn't have data at the time. Spawn, Spawn Type, Paramaters, and State can all be number fields. We also don't need to create any extra functions for parsing them. Just doing:
```cpp
const int stage = entry["Stage"].as<int>();
const int room = entry["Room"].as<int>();
const int spawn = entry["Spawn"].as<int>();
const int layer = entry["State"].as<int>();
```
is preferable to me since I'd rather have it throw an error when generating if the data is missing/formatted incorrectly, rather than silently keep going with invalid data.


As far as no logic goes, this implementation seems fine, but I'd prefer if we check for no logic in `ParseRequirementString` and just return `NO_REQUIREMENT` if no logic is enabled. This way we aren't performing unnecessary logic checks in no logic when running the flattening algorithm.

JaxonKEKW:
Ahh okay. I didn't really have an idea of the approach you had in mind, thank you. I'm gonna convert to a draft and work on this a bit later, starting with the entrance stuff and try to get it working the way you thought up

PR #1828, opened 2026-05-26
Scale Memory Card at Widescreen, fix Name Select screen

JaxonKEKW:
- Scales the Memory Card screen (text only) with Wii and Dusklight menu scaling settings (GC already keeps default scaling)

defaultMem
scaledMem

- Fix oversight on the Name select screen with GC menu scaling settings (all text was getting scaled down)

defaultName
scaledName

- Deleted a comment I left in and moved some panes to header files because they can just be looped through with others. Also initialized all variables for each PaneCache, just realized I was only initializing tag, origX and the bool previously (originally only had origX and added origY later, forgot to adjust) 

- All new logic and definitions are in TARGET_PC ifdefs, and moved ones were also previously.


Irastris:
Had a few formatting nits but GitHub again refuses to let me push into your branch, so here's a patch if you could please apply it.

[nits.patch](https://github.com/user-attachments/files/28271965/nits.patch)

JaxonKEKW:
patched the formatting, sorry about that

Let me know if there's anything else you see

Irastris:
Just the review I posted asking about cachedPanesName, once I have clarification on that it'll be good to merge IMO

JaxonKEKW:
Oh sorry I completely forgot I did that. Since file_select.cpp includes d_name.h, both files see the cachedPanes variable and I just set that up to test changes and never went back to do it properly. I’ll change that when I get home

JaxonKEKW:
Moved cachedPanes to the functions, could honestly probably just move everything back into the funcs like the original code has it, but didn't wanna have the wide funcs be longer than they needed to be, let me know

src/d/d_name.cpp
```diff
@@ -1430,51 +1430,86 @@ void dName_c::selectCursorPosSet(int row) {
...
+                entry.origTransY = pane->getTranslateY();
+                entry.cached = true;
+            }
+        }
+        for (PaneCache& entry : l_nameCurTagName) {
+            J2DPane* pane = nameIn.NameInScr->search(entry.tag);
+            if (!entry.cached) {
+                entry.origTransX = pane->getTranslateX(); 
+                entry.origTransY = pane->getTranslateY();
+                entry.cached = true;
```
Irastris: Same here

encounter (approved):

PR #1841, opened 2026-05-26
Add option for inverting Direct Select

Abzol:
Adds a new subheading under Input for gameplay-related settings. It's at the bottom because players probably will want to configure their camera more than they care about ornery gameplay details.

The one and only toggle there right now is to invert the function of the L-trigger on the Item Wheel. It renders the on-screen text properly if inverted too, meaning with it on the game does tell you that holding L will rotate and moving the stick without holding L will direct select.


encounter (approved):
Works well, thanks!

PR #1857, opened 2026-05-27
Fix mSkipEventName again

### LINKED ISSUE #1531: Game crash when talking to Zora at the bottom of the waterfall in Zora's Domain.
When you go to fast travel to the top of the waterfall the game crashes. Tested multiple times.

I'm on Linux appimage version 1.1.1

On SteamOS (Steam Deck)

I'm at the point of the story immediately after beating the Gerudo Desert dungeon.

PJB3005:
They WAY undersized this in the original game it seems. The event used for the Zora overflows the original buffer size by *6* bytes.

Fixes https://github.com/TwilitRealm/dusklight/issues/1531

### DIFF
--- a/include/d/d_event.h
+++ b/include/d/d_event.h
@@ -196,8 +196,8 @@ public:
     /* 0x108 */ int mSkipTimer;
     /* 0x10C */ int mSkipParameter;
     /* 0x110 */ BOOL mIsSkipFade;
-#if TARGET_PC
-    /* 0x114 */ char mSkipEventName[21];
+#if AVOID_UB
+    /* 0x114 */ char mSkipEventName[32];
 #else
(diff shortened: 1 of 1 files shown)


Krutonium:
Weird; I actually tested the waterfall after my original fix and didn't have any issues. I wonder if it uses a different cutscene if some other flag is set

tomlube:
It only crashed at night for whatever reason

Krutonium:
Maybe it uses a different cutscene entirely at night? Seems a bit arbitrary, but w/e if this fixes it lol. Prior to my fix it crashed in the day as well.

PJB3005:
> Maybe it uses a different cutscene entirely at night?

The event name is different yes, the extra `_NIGHT` in the name is part of the problem.

PR #1864, opened 2026-05-27
Add a *ton* of const and constexpr around the codebase.

PJB3005:
This makes the codebase compile without strings being cast to non-const char*. I also went through and added constexpr where appropriate for tons of static data.


GiorgioMendieta:
Great, this fixes a lot of the warnings that were appearing during build

TakaRikka (approved):
think this is fine, but its a massive change so would like another set of eyes

PR #1865, opened 2026-05-27
fix: change enums from int to u8

GiorgioMendieta:
title says it all

### DIFF
--- a/include/dusk/settings.h
+++ b/include/dusk/settings.h
@@ -9,19 +9,19 @@ namespace dusk {
 
 using namespace config;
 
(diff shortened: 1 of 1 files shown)


encounter:
Don't think this is a good idea for these enums

GiorgioMendieta:
> Don't think this is a good idea for these enums

Why is that? Other enums in that file are u8 as well, it uses 1 byte compared to the usual 4 for ints. 

Besides, I don't think there would be more than 255 options so it would benefit from saving up on memory even if it's a small gain no? 

Please let me know what you think, thanks :)

PR #1869, opened 2026-05-27
feat: Change Menu Scaling option to graphics_tuner_control

GiorgioMendieta:
Similar to #1599 

Lets users preview the menu scaling option without the settings pane obstructing the view

# Before

https://github.com/user-attachments/assets/cbd3ba8d-5b71-4e49-816e-bda0b7aeb935

# After

https://github.com/user-attachments/assets/40c7a9bd-c3ea-4701-a3ec-77454a4f94f5

### DIFF
--- a/include/dusk/settings.h
+++ b/include/dusk/settings.h
@@ -54,7 +54,7 @@ enum class FrameInterpMode : u8 {
 enum class MenuScaling : u8 {
     GameCube = 0,
     Wii = 1,
-    Dusklight = 2,
+    Dusk = 2,
 };
 
 namespace config {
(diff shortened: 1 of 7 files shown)


Abzol:
I think this will raise more questions for users than it answers when they look at a gameplay scene and try to figure out whats changing

GiorgioMendieta:
> I think this will raise more questions for users than it answers when they look at a gameplay scene and try to figure out whats changing

I don't think so. It specifically says it affects **Collection and File sharing menus** on the setting. Even if this PR doesn't get merged you could argue the same thing with its current implementation.

I personally think if users are modifying the settings, they should be expected to read what the setting specifically does

encounter:
I agree with Abzol that this may be more confusing than beneficial

PR #1871, opened 2026-05-27
Set daObjAvalanche_c Matrix Properly

### LINKED ISSUE #1838: [1.3.0][Linux x86_64] Snow Ramp Animation not triggering
The Animation for the snow ramp is not triggering when i head bump into the wall. 
Ive tried it with attacking and running.
After switching back to version 1.2.0 it worked again.

Image

SuperDude88:
Fixes #1838


TakaRikka:
is this an og game bug or something?

SuperDude88:
> is this an og game bug or something?

We (Olivia/Jasper/I) assume so based on the code + inconsistent behavior, Jasper said that `the intent is that it starts with S (to set) and things after use M (to multiply)` for mDoMtx_stack_c operations (which the original code here does not adhere to)

PR #1896, opened 2026-05-28
fix: item and talk rendering

Sirius902:
Fixes some UI elements not rendering. I guess #1864 missed some stuff so these methods are no longer overriding anything since the prototype doesn't match so it just uses the virtual function which does nothing. Maybe it'd be worth adding `override` all over the codebase like the `const`/`contexpr` changes so this is a static compile error.

### Before

Midna Text Invisible

### After

Midna Text Visible


PJB3005:
Ugh, what a C++ moment. Thanks for the fix.

PJB3005 (approved):

PR #1909, opened 2026-05-29
Added Sword and Shield skip Achievement

5upern1ce:
Redid the pull request to refactor my repository because I totally misunderstood how forking worked, sorry about that guys. 

If both the Ordon Sword and Shield aren't in the inventory when activating the flag for Midna letting Link through the Twilight Wall (M_013) the achivement is given.

PR #1917, opened 2026-05-29
Revert "Added Sword and Shield skip Achievement"

tomlube:
Reverts TwilitRealm/dusklight#1909

### DIFF
--- a/src/dusk/achievements.cpp
+++ b/src/dusk/achievements.cpp
@@ -933,29 +933,6 @@ std::vector<AchievementSystem::Entry> AchievementSystem::makeEntries() {
         },
         {
             {
-                 "sword_and_shield_skip", "Sword and Shield Skip",
-                 "Pass through the Twilight Wall in Ordon Spring before obtaining both the Ordon Sword "
-                 "and Shield.",
-                 AchievementCategory::Glitched, false, 0, 0, false
-            },
-            [](Achievement& a, json&) {
-                bool hasOrdonSword = false;
-                bool hasOrdonShield = false;
-                if (dComIfGs_isCollectSword(COLLECT_ORDON_SWORD)) {
-                    hasOrdonSword = true;
-                }
-                if (dComIfGs_isCollectShield(COLLECT_ORDON_SHIELD)) {
-                    hasOrdonShield = true;
-                }
-                if ((!hasOrdonSword || !hasOrdonShield) &&
-                    dComIfGs_isEventBit(dSv_event_flag_c::M_013)) {
-                    a.progress = 1;
-                }
-            },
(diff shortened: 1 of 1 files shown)


Irastris (approved):

5upern1ce:
Out of curiosity, why was this reverted? Was there an issue I didn't notice?

src/dusk/achievements.cpp
```diff
@@ -931,29 +931,6 @@ std::vector<AchievementSystem::Entry> AchievementSystem::makeEntries() {
...
-            [](Achievement& a, json&) {
-                bool hasOrdonSword = false;
-                bool hasOrdonShield = false;
-                if (dComIfGs_isCollectSword(COLLECT_ORDON_SWORD)) {
-                    hasOrdonSword = true;
-                }
-                if (dComIfGs_isCollectShield(COLLECT_ORDON_SHIELD)) {
-                    hasOrdonShield = true;
-                }
-                if ((!hasOrdonSword || !hasOrdonShield) &&
```
Abzol, line 950: Should be `!hasOrdonSword && !hasOrdonShield`

Irastris:
Thought Tom would have dropped a comment on the PR, sorry.

It was merged by accident while we were discussing its logic and presentation internally. There's the above review from Abzol, and we also feel that it needs a cleverer name.

src/dusk/achievements.cpp
5upern1ce: I think that BiT gives the Ordon sword but not the shield, which would make sword and shield skip in the any% route not pop the achievement no? Or does the fact that the sword doesn't show on the inventory screen mean that shouldn't be true?

5upern1ce:
Oh yeah for sure I get the cleverer name, I'm working on some other achievements because I really like that you guys implemented them into this port, so I'm gonna really challenge myself to come up with some clever names!! Thanks for the feedback I appreciate you taking the time to do that

tomlube:
Sorry, I meant to comment and got caught up with life stuff instead. 

Name could be "Hugo Leaving"

5upern1ce:
Haha that's a good one! I love that the community decided he needed a name

tomlube:
https://docs.google.com/spreadsheets/d/1qCjA9_u3HK5qQByTHuH24dQoX2f09Y6YdF7MWSy4Z9Y/edit?gid=0#gid=0

Very important document by the way

5upern1ce:
Oh my god bro that's so peak, I was watching bewildebeest practicing all dungeons Lakebed and so I know purplefriend and everyone knows Hugo but now I have so many names to learn! yippee!!

PR #1927, opened 2026-05-30
Revert "Adds a new cheat that let you transform from the start of the game."

encounter:
Reverts TwilitRealm/dusklight#1241

Sadly it breaks stuff


Abzol:
Fair. Ive also come to think it probably would do better as a code mod in the future.

Issue #1883, opened 2026-05-28
invincible link request


spkpal:
### Feature Description

This request would help out alot people incase they want to practice the Stallord skip and stuff, it would help out alot of people on there speedruns as well, so can you think about adding this request please

Abzol:
duplicate of #1861

Issue #1861, opened 2026-05-27
Feature request, Invincible disable link hurt box


spkpal:
I would like to request a Invincible cheat that disables link hurtbox please

MelonSpeedruns:
We already have an infinite hearts option. If you want to simply not have Link play the damage animation, then this sounds like something you could implement when our Modding API comes out!

spkpal:
@MelonSpeedruns please understand i have trouble playing games thats why i need extra help on them and this is my favourite game of all time

thats why i wanted to request this cheat so would it help me play my game a whole alot more better for me

spkpal:
this is the reason why i need that cheat to be added please https://www.youtube.com/watch?v=6lUVMAx1itc

because i have trouble flying the bird up the river

spicyjuice04:
> We already have an infinite hearts option. If you want to simply not have Link play the damage animation, then this sounds like something you could implement when our Modding API comes out!

To be fair, speedrunners have often used this feature in tpgz especially for practice on Stallord Skip. The difference between infinite hearts and the invincible cheat is that your position can't be messed up. Though this post was probably not made originally with that in mind, just a nice to have. Otherwise people would probably mod it in themselves like you suggested.

spkpal:
I love it when my position can't be messed up with the invincible cheat, because it helps me with the flying bird up the river and it helps me with the Stallord boss because of the spike beyblades on the railing so my position can't be messed up

PR #1943, opened 2026-05-31
Invincible Link

xCyberxx:
As requested in https://github.com/TwilitRealm/dusklight/issues/1883 and https://github.com/TwilitRealm/dusklight/issues/1861. These changes introduce an invincible mode for Link. Enabling this will result in damage not being registered and Link not getting stunned or receiving knockback. 

image

### DIFF
--- a/include/dusk/settings.h
+++ b/include/dusk/settings.h
@@ -230,6 +230,7 @@ struct UserSettings {
         ConfigVar<bool> fastSpinner;
         ConfigVar<bool> freeMagicArmor;
         ConfigVar<bool> invincibleEnemies;
+        ConfigVar<bool> invincibleLink;
 
         // Technical
         ConfigVar<bool> restoreWiiGlitches;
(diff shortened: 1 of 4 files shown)


SuperDude88:
Would the use cases for this generally be covered by cheating in magic armor and setting it to not use rupees?

spicyjuice04:
> Would the use cases for this generally be covered by cheating in magic armor and setting it to not use rupees?

Yea, I looked at the files changed. The request in https://github.com/TwilitRealm/dusklight/issues/1861. and https://github.com/TwilitRealm/dusklight/issues/1883 were about the complete interaction of taking knockback, similar to what the tpgz cheat does. The files changed here just make you not take damage but the knockback still happens. The way it is implemented here is just a different step as the infinite hearts cheat

xCyberxx:
No, with these changes Link basically gets permanently put in invincibility frames, which cancels any reaction to, for example, hits from arrows as wolf, knockback (to the ground) from helmasaurs, etc. Which is significantly different from how infinite hearts or how magic armour reacts to these situations.

You can build this branch to play around and see how it reacts

Korados:
Why doesn't this merge into the Windows builds? I would've loved to test this. Just a heads up with that cheat: When Link is fully invincible, he also can't recieve statuses, like getting buffed by the Great Fairy Tears or Rare Chu Jelly. At least, this is how it is with the TPHD Cheat Engine table.

PR #1945, opened 2026-05-31
Crash Reporting Option Portability

SuperDude88:
Still has details I need to iron out. Goal is to preserve the crash report setting if people change devices with a portable install.


encounter:
Initially I was trying to avoid putting sentry stuff in the data folder so it's not sync'd. Theoretically we could store it in the config and then just tell sentry it was consented if it stops knowing that

Issue #1862, opened 2026-05-27
Flatpak Tracking Issue


ReillyBrogan:
This issue exists to track the remaining tasks ahead of getting Dusklight building as a Flatpak as well as in Flathub for user consumption.

### Nod
- [x] Add vendored crates for offline builds [[PR Opened](https://github.com/encounter/nod/pull/19)]
- [ ] Use system versions of liblzma, libzstd, libbz2, and libz [[Issue](https://github.com/encounter/nod/issues/20)] (Note that this is a nice to have and will mainly reduce binary size and build times)

### Dawn
- [ ] Build Dawn source archive for offline builds [[Issue](https://github.com/encounter/dawn-build/issues/3)] (Nice-to-have)

### Aurora
- [x] Add ability to override resource directory location [[PR Opened](https://github.com/encounter/aurora/pull/211)]
- [x] Use runtime-provided libsqlite3 and libzstd [[Issue](https://github.com/encounter/aurora/issues/212)] (Nice-to-have)
- [ ] Manage fetchcontent dependencies centrally [[Issu](https://github.com/encounter/aurora/issues/213)] (Nice-to-have)

### Dusklight
- [x] Add metainfo file [[PR Opened](https://github.com/TwilitRealm/dusklight/pull/1860)]
- [x] Install dusklight binary to bindir and resources to datarootdir (I have this done locally but waiting on merging the Aurora PR)
- [ ] Install freedesktop assets (icons/desktop/metainfo) using cmake directly
- [x] Ensure that Aurora uses runtime libzstd (Nice-to-have)
- [ ] Handle pre-cached pipeline shaders
- [ ] Manage fetchcontent dependencies centrally (Nice-to-have)

### Flatpak
- [ ] Add to repo
- [ ] Submit to Flathub (Note that we can do this early before everything else is done)
- [ ] Automate release notes generation
- [ ] Automate dependency management
- [ ] Integrate into Github CI
- [ ] Add to Github releases workflow
- [ ] Add pipeline to push new releases to Flathub

## Testing
- I have noticed that the Appimage currently defaults to launching via xwayland/x11 (unsure why). The Flatpak correctly uses Wayland by default when available which is something to keep in mind.
- [ ] Does moving the data directory work or does that need to be disabled?
- [ ] Do we want to add back the update checker but make it work via Flathub APIs somehow?
- [ ] I've tested that Discord rich presence works with native Discord/Vesktop. Does it work with Discord installed via Flatpak?

## Documentation
[...]

PR #2023, opened 2026-06-08
Add Flatpak

ReillyBrogan:
- Change appId on Linux to dev.twilitrealm.Dusklight
- Install assets with cmake
- Fix platform name (was getting set to "cmake_system_name")
- Add flatpak

Part of https://github.com/TwilitRealm/dusklight/issues/1862


CMakeLists.txt
encounter: Maybe we conditionally set this to `.Dusklight` for Linux builds and use this variable downstream instead of `LINUX_APP_ID`. We can also include it in our `version.h.in` so that `SDL_SetAppMetadata` can use it directly rather than the conditional logic

ReillyBrogan: I've taken that a little bit further. The app_id is now configurable and defaults to `dev.twilitrealm.Dusklight` on Linux and `dev.twilitrealm.dusk` on all other platforms. With `DUSK_PACKAGE_INSTALL` all files installed use the app_id as part of their name or install location and so it should be possible for forks to change the app_id to be co-installable without any further change.

andreas-traexler:
Any news about this?

PR #2025, opened 2026-06-08
Add setting to memorize window size

Kyrio:
Until now, Dusklight would either start at a window size of twice the "default resolution" of 608x448, or start in fullscreen if it had been toggled in the last session. With this new setting, resizing the window will immediately save the new window size in settings, so that it can be restored at the next startup.

I added this mostly as a way to get familiar with the codebase, feedback is appreciated!

image

### DIFF
--- a/src/m_Do/m_Do_main.cpp
+++ b/src/m_Do/m_Do_main.cpp
@@ -260,7 +260,7 @@ void main01(void) {
                 dusk::g_imguiConsole.HandleSDLEvent(event->sdl);
                 break;
             case AURORA_WINDOW_RESIZED:
-                if (dusk::getSettings().video.memorizeWindowSize && !dusk::getSettings().video.enableFullscreen) {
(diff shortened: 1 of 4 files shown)


Irastris (changes requested):
Disabling the option does not restore the original behavior of Dusklight creating its window at 2x the native framebuffer size, instead it keeps using whatever the last saved window size was.

src/m_Do/m_Do_main.cpp
```diff
@@ -586,6 +593,17 @@ int game_main(int argc, char* argv[]) {
...
+        if (dusk::getSettings().video.memorizeWindowSize) {
+            int width = dusk::getSettings().video.lastWindowWidth.getValue();
+            int height = dusk::getSettings().video.lastWindowHeight.getValue();
+
+            if (width > 0 && height > 0) {
+                config.windowWidth = width;
+                config.windowHeight = height;
+            }
+        }
+
```
Irastris, line 605: Not a fan of this syntax, just set the window size once depending on the outcome

Irastris (approved):

PR #2034, opened 2026-06-11
monkey crash fix

matthewdavidrichardanderson:
maybe dont merge this because idk what im doing and i had to use a lot of ai, but this has fixed all the crashes speedrunners run into in forest temple from my testing. logs showed the crash consistently happening after a monkey loaded its last resource and this seems to have fixed it, thats all i really know, im not sure why it crashed more often with ebf


src/d/actor/d_a_npc_ks.cpp
```diff
@@ -6882,6 +6882,10 @@ static int daNpc_Ks_Delete(npc_ks_class* i_this) {
         i_this->model->stopZelAnime();
     }
 
+    if (saru_p[i_this->set_id] == i_this) {
```
encounter, line 6890: Please wrap this in `#if TARGET_PC` / `#endif`

matthewdavidrichardanderson: done

encounter (approved):
Thanks, verified this is a good fix and made it clear `leader` as well

PR #2075, opened 2026-06-16
Add setting to disable targeting letterbox

### LINKED ISSUE #1369: [Suggestion] Add the option to remove the black bars when targeting with Z
It's something that bothers me, and I think it might bother someone who isn't used to it even more. If it were possible to remove it, I would be very happy.

samkleespies:
Adds a setting to disable the top and bottom letterbox black bars when L-targeting / holding L.

Closes #1369.

### DIFF
--- a/src/d/d_camera.cpp
+++ b/src/d/d_camera.cpp
@@ -2721,6 +2721,12 @@ int dCamera_c::defaultTriming() {
         case 2:
         case 5:
         case 6:
+#if TARGET_PC
+            if (dusk::getSettings().game.disableTargetingLetterbox) {
+                SetTrimSize(0);
+                break;
+            }
+#endif
             SetTrimSize(1);
             break;
         case 4:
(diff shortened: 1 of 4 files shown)


MelonSpeedruns (approved):
Looks good to me

src/d/d_camera.cpp
baxlek, line 2712: Adding the check here using an IF_DUSK(|| allows for disabling letterboxing entirely if that would be prefered.

MelonSpeedruns:
As suggested by @baxlek , I think we'd rather have such a feature if it's global instead of just for L-targeting.
Maybe make it a 4-way option like the Magic Armor, Where people can have them: `On - Off - Only during Cutscenes - Only during Gameplay`.

baxlek:
I went ahead and made a [build](https://github.com/TwilitRealm/dusklight/compare/main...baxlek:dusklight:disable-letterboxing-options) with all the options that @MelonSpeedruns suggested. However, I'm unsure of how to proceed. For one, this PR is still open and approved. Second, I relied heavily on Copilot. Mostly to understand the codebase and how to implement the 4-way option.

Issue #1077, opened 2026-05-12
Save slot times don't match system time


DanKorell:
I noticed last night that when saving the game, my save slot had a time one hour behind local time. It looks like it might be using UTC time instead of local time.

PR #2089, opened 2026-06-17
System Time Sync

### LINKED ISSUE #1919: Time Sync
### Feature Description

Synchronize Hyrule's time scale with real-world time flow. Essentially a port of the "Ship of Harkinian" feature of the same name.

### Platform

_No response_

### Additional Information

Previous feature request (#1656) was erroneously closed as a duplicate of an unrelated issue.

ZipoLabs:
Adds a cheat, "System Time Sync", to synchronise the in-game time of day to the system clock, similarly to in Ship of Harkinian.

Closes #1919


MelonSpeedruns (dismissed):
Looks good to me! This looks like a feature we definitely want in Dusklight. Maybe we could have a weather sync next as part of another PR? 👀

MelonSpeedruns:
One bug I noticed, the editor says 7:49 AM, even though my PC is at 8:49 AM. Do you account for daylight savings?

ZipoLabs:
The daylight savings bug appears to be an existing issue with Dusklight, also affecting the save time. I had hoped to investigate a solution in a later, separate PR.

This has previously been brought up in #1077.

MelonSpeedruns:
Oh good catch! I can see about fixing the daylight saving bug in dusklight first, then we can merge this PR. You can also check about fixing it on your side if you want.

MelonSpeedruns:
https://github.com/encounter/aurora/pull/229

Can confirm this fix I just made fixes the file select and this PR being 1 hour behind.

Abzol:
Some notes:
 - I'm not sure this should be a cheat? I think it's more suited to be under the "Gameplay" options.
 - It should be disabled in speedrun mode.
 - "Breaks" Sun's Song in the same way as howling in a zone where time doesn't pass; maybe it should call up the same "Nothing happened..." text?
 - This breaks certain scripted time events; Most easily reproduceable are exiting City in the Sky via the cannon (sets ingame time to 12:00 no matter what, not very important) and entering Stallord Arena (sets time to 21:00 because the scene in Mirror Chamber after the boss is supposed to take place at night). Stallord Arena can be easily reproduced by warping to Arbiter's Grounds Room 9 Point 1, City in the Sky Room 0 Point 0 is fine.

baxlek (changes requested):
I went ahead and fixed the sleep mode desync by using the actual device clock instead of the "steady clock" aurora uses. Now also allows for setting arbitrary time by changing the device's time.

Also fixed an issue in the fishing hole where time would "run away" endlessly during certain hours.

[d_kankyo.zip](https://github.com/user-attachments/files/30173624/d_kankyo.zip)

baxlek (commented):
Do whatever.

baxlek (approved):

tomlube:
closing since it's now a great mod available in the mod menu :) thank you for the effort!

PR #2150, opened 2026-06-29
Add command line arguments for --develop, --load-save, and --stage

jdflyer:
This PR adds 3 new command line arguments to dusklight that are designed to speed up development.

- --develop: Sets mDoMain::developmentMode to true and enables dusk::OSReportReallyForceEnable (The same behavior as the original development mode)
- --load-save launches dusklight into a specified save from the existing savefile. It overrides the scene change function in d_s_logo and loads the save and goes straight to the play scene
- --stage launches dusklight into a specified stage with the following format: STAGE,ROOM,POINT,LAYER with the last few parameters being optional. This is implemented in the same function as --load-save and the two work together.


src/d/d_s_logo.cpp
```diff
@@ -757,7 +762,80 @@ void dScnLogo_c::nextSceneChange() {
...
+                if (dusk::SaveRequested == 0) {
+                    dComIfGs_init();
+
+                    fopScnM_ChangeReq(this, fpcNm_PLAY_SCENE_e, 0, 30);
+                    dusk::SaveRequested = 0xff; //Skip requesting the scene from above
+                }
+
+                // Use both to force-set start stage
+                dComIfGp_setNextStage(stageName.c_str(), point, room, layer);
+                g_dComIfG_gameInfo.play.mNextStage.getStartStage()->set(stageName.c_str(),point,room,layer);
```
encounter: `room` and `point` are reversed, it seems

src/d/d_s_logo.cpp
```diff
@@ -757,7 +762,80 @@ void dScnLogo_c::nextSceneChange() {
...
+                std::stringstream ss(dusk::StageRequested);
+                std::string token;
+
+                std::getline(ss,token,',');
+                std::string stageName = token;
+                s8 room = 0;
+                s16 point = 0;
+                s8 layer = -1;
+                if (std::getline(ss,token,',')) {
+                    room = std::stoi(token);
```
encounter: `std::stoi` can throw `std::invalid_argument` or `std::out_of_range`, would be nice to handle those gracefully and print usage info

jdflyer: Thanks, I moved all the parsing of the stage into main so any errors can be caught before dusklight launches.

src/d/d_s_logo.cpp
jdflyer: Fixed :)

src/m_Do/m_Do_main.cpp
```diff
@@ -534,11 +533,52 @@ int game_main(int argc, char* argv[]) {
...
+    }
+
+    if (parsed_arg_options.contains("load-save")){
+        uint8_t slot = parsed_arg_options["load-save"].as<uint8_t>();
+        if (slot >= 1 && slot <= 3) {
+            dusk::SaveRequested = slot;
+        }
+    }
+
+    dusk::registerSettings();
```
jdflyer, line 581: Note: I also moved this down just to fix a bug that crashed while unregistering settings if we exited while parsing arguments or displaying help.

Issue #2059, opened 2026-06-15
Add fullscreen monitor resolution option


ggmoritz:
### Feature Description

Currently, the rendering resolution can already be changed.
Many games provide a separate option to override the monitors resolution while in fullscreen though, thus enabling use of DLDSR without needing to fiddle with the display settings every time.
This would be especially useful because the game has no anti aliasing and is very easy to run at a higher resolution than native

### Platform

Windows/PC

Irastris:
I'm not entirely sure this would provide a better experience over the existing solution of raising your internal resolution and using the Area sampler, which together already provide pretty good AA, but leaving the issue open for the time being

PR #2187, opened 2026-07-08
Add "Fullscreen Resolution" option

ggmoritz:
Adds Video > Display dropdowns to override the monitor resolution and refresh rate while in exclusive fullscreen.
Enables using NVIDIA DSR / AMD VSR to run above native resolution without changing Windows display settings.
"Desktop" keeps the existing borderless-fullscreen behavior.
Implements my [pull request](https://github.com/TwilitRealm/dusklight/issues/2059)

### DIFF
--- a/files.cmake
+++ b/files.cmake
@@ -1439,6 +1439,8 @@ set(DUSK_FILES
         include/dusk/menu_pointer.h
         src/dusk/menu_pointer.cpp
         src/dusk/mouse.cpp
+        include/dusk/display.h
+        src/dusk/display.cpp
 		src/dusk/gamepad_color.cpp
 		src/dusk/autosave.cpp
         src/dusk/http/http.hpp
--- a/include/dusk/display.h
+++ b/include/dusk/display.h
@@ -1,0 +1,24 @@
+#pragma once
+
+#include <vector>
+
+namespace dusk::display {
+
+struct Resolution {
+    int width;
+    int height;
+};
+
+// Distinct fullscreen resolutions offered by the display the window is currently on, sorted from
+// largest to smallest. Includes driver-injected modes (e.g. NVIDIA DSR/DLDSR) when enabled.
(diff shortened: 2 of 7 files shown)


Calinou:
> Enables using NVIDIA DSR / AMD VSR to run above native resolution without changing Windows display settings.

If Dusklight supports setting an internal resolution scale, it could be set above 100% to remove the need to use DSR/VSR and have something that also works on non-Windows platforms, or in windowed mode.

Ultimately, DSR/VSR come with many downsides, so in-game support for supersampling is preferred instead.

ggmoritz:
> > Enables using NVIDIA DSR / AMD VSR to run above native resolution without changing Windows display settings.
> 
> If Dusklight supports setting an internal resolution scale, it could be set above 100% to remove the need to use DSR/VSR and have something that also works on non-Windows platforms, or in windowed mode.
> 
> Ultimately, DSR/VSR come with many downsides, so in-game support for supersampling is preferred instead.

Perhaps for AMD an internal resolution scale would be better but on NVidia DSR includes DLDSR (deep learning DSR), which produces a superior image at the equivalent resolution.

Calinou:
> NVidia DSR includes DLDSR (deep learning DSR), which produces a superior image at the equivalent resolution.

DLDSR is just about improved downsampling, it can't improve temporal stability (since it doesn't add subpixel detail on its own). It might look better in stills, but stability in motion will be better accomplished using traditional supersampling.

ggmoritz:
> > NVidia DSR includes DLDSR (deep learning DSR), which produces a superior image at the equivalent resolution.
> 
> DLDSR is just about improved downsampling, it can't improve temporal stability (since it doesn't add subpixel detail on its own). It might look better in stills, but stability in motion will be better accomplished using traditional supersampling.

That is true but doesn't contradict my point.

SuperDude88:
This isn't something we currently plan to support/isn't how we would want to implement this

PR #2206, opened 2026-07-13
Draw shadows before translucent textures

GiorgioMendieta:
Draws shadows before translucent overlays, makes the scene look cleaner. This is more noticeable when the shadow strength is higher. 

This PR does not include the debug on/off toggle, it was just to show the difference

https://github.com/user-attachments/assets/3ca4f066-5a74-4702-b926-d5247852156a

*Comparison*
Before:
image

After:
image


encounter:
Thanks!

PR #2220, opened 2026-07-17
Nix: add miniz and fix submodules

jack-wines:
Hi, and thank you for this wonderful port. I was building locally and noticed that nix build wasn't working because of `miniz`. I added it in so that it compiles and runs now. Happy to report that nix and cmake's builds both work on my arm linux machine.

Also, nix should now include submodules without issue.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -59,21 +59,6 @@
           hasNodPrebuilt = nodPrebuiltInfo ? ${system};
 
           aurora = builtins.pathExists "${self}/extern/aurora/CMakeLists.txt";
-          needSubmodules = ''
-            dusklight: The aurora submodule is not vendored. Add submodules=1 to build.
-
-            As a flake input:
-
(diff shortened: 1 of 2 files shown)


flake.nix
qubitnano: With inputs.self.submodules = true; this won't be needed anymore

flake.nix
```diff
@@ -169,6 +178,7 @@
               throw needSubmodules
             else
               pkgs.stdenv.mkDerivation {
+                enableParallelBuilding = true;
```
qubitnano: cmake in nixpkgs should enableParallelBuilding by default

flake.nix
```diff
@@ -169,6 +178,7 @@
               throw needSubmodules
             else
```
qubitnano: This won't be needed anymore with inputs.self.submodules = true;

jack-wines:
Agreed and fixed on all three counts.

encounter:
Thanks!

jack-wines:
you are most welcome!

PR #2226, opened 2026-07-17
Button Fishing Improvement

ZipoLabs:
Adds compatibility for the "Button Fishing" gameplay setting when casting from the canoe in the Fishing Pond.


encounter (approved):
Thank you!

PR #2227, opened 2026-07-18
japanese gamecube version support

TakaRikka:
thanks to @pengin for figuring out the font rendering fix


src/d/d_msg_class.cpp
```diff
@@ -2108,8 +2150,8 @@ bool jmessage_tSequenceProcessor::do_isReady() {
             }
 
             field_0xa6++;
-#if REGION_JPN
-            if (field_0xa6 >= 1) {
+#if TARGET_PC || REGION_JPN
+            if (IF_DUSK(dusk::version::isRegionJpn() &&) field_0xa6 >= 1) {
```
encounter: This logic is written wrong (original checked >=2)

src/d/d_msg_class.cpp
```diff
@@ -1591,8 +1630,8 @@ void jmessage_tMeasureProcessor::do_scale(f32 i_scale) {
         mPageLineMax--;
         JUT_ASSERT(0x930, mPageLineMax > 0);
 
-#if REGION_JPN
-        if (field_0x3e == 0) {
+#if TARGET_PC || REGION_JPN
+        if (IF_DUSK(dusk::version::isRegionJpn() &&) field_0x3e == 0) {
```
encounter: This also loses the original non-JPN logic

src/Z2AudioLib/Z2SoundObjMgr.cpp
```diff
@@ -207,6 +209,10 @@ void Z2SoundObjMgr::searchEnemy() {
         case Z2_ENEMY_RD:
         case Z2_ENEMY_SH:
         case Z2_ENEMY_HP:
+#elif TARGET_PC
+        case Z2_ENEMY_SH:
+            if (enemyId == Z2_ENEMY_SH && !dusk::version::isRegionJpn())
```
encounter, line 221: This logic doesn't look right, I think it'd be closer to:
```c++
case Z2_ENEMY_NZ:
case Z2_ENEMY_PO:
case Z2_ENEMY_HP:
    isGhostEnemy = true;
#if TARGET_PC
    if (enemyId == Z2_ENEMY_HP && dusk::version::isRegionJpn() &&
        Z2GetLink()->isRiding() && Z2GetLink()->getMoveSpeed() > 38)
    {
        continue;
    }
#endif
    break;

// ...

#elif TARGET_PC
case Z2_ENEMY_SH:
    if (enemyId == Z2_ENEMY_SH && !dusk::version::isRegionJpn())
        break;
#endif
    if (!Z2GetLink()->isRiding())
        break;
    if (Z2GetLink()->getMoveSpeed() <= 38)
        break;
    continue;
```

libs/JSystem/src/JUtility/JUTResFont.cpp
```diff
@@ -273,12 +283,19 @@ f32 JUTResFont::drawChar_scale(f32 pos_x, f32 pos_y, f32 scale_x, f32 scale_y, i
 
     u16 texW  = mpGlyphBlocks[field_0x66]->textureWidth;
 #if TARGET_PC
-    u16 texH  = mJoinedTextureHeight;
+    // JUTCacheFont does not set mJoinedTextureHeight (it uses per-page textures via loadImage override).
+    // Fall back to the individual glyph block's textureHeight in that case.
+    u16 texH  = mJoinedTextureHeight > 0 ? (u16)mJoinedTextureHeight : (u16)mpGlyphBlocks[field_0x66]->textureHeight;
```
encounter, line 288: It looks like we never call `initJoinedTexture` from `JUTCache` in particular, so we need to zero `mJoinedTextureHeight` in `initialize_state` otherwise it's uninitialized memory

src/d/d_menu_dmap.cpp
```diff
@@ -429,13 +477,22 @@ static f32 player_py;
 DUSK_GAME_DATA dMenu_Dmap_c* dMenu_Dmap_c::myclass;
 
 void dMenu_DmapBg_c::setCButtonString(u32 i_msgNo) {
+#if TARGET_PC
+    static u64 const c_tag_jpn[2] = {
+        MULTI_CHAR('c_text_s'), MULTI_CHAR('c_text')
+    };
+    static u64 const c_tag[2] = {
+        MULTI_CHAR('c_text_s'), MULTI_CHAR('c_text')
```
encounter:
```suggestion
        MULTI_CHAR('f_text_s'), MULTI_CHAR('f_text')
```

src/d/d_msg_class.cpp
```diff
@@ -3368,13 +3443,24 @@ void jmessage_tRenderingProcessor::do_heightcenter() {
...
         }
         break;
     case 3: {
-        if (field_0x142 == 1) {
+#if TARGET_PC
+        if (!dusk::version::isRegionJpn() || (dusk::version::isRegionJpn() && field_0x142 == 1))
+#else
+        if (field_0x142 == 1)
+#endif
+        {
```
encounter:
```suggestion
#if TARGET_PC
        if (!dusk::version::isRegionJpn()) {
            field_0x138 = 0.5f * pReference->getLineSpace();
            var_f31 += field_0x138;
            break;
        }
#endif
        if (field_0x142 == 1) {
```
This matches the original case 3 logic below (no `pReference->getLineMax() - (s16)nowPageLine`)

encounter (approved):

PR #2230, opened 2026-07-19
Mod SDK: Game Mode Service

jdflyer:
This PR adds a Gamemode Service to dusklight. It allows a gamemode to be registered that allows the game to designate one specific form of gameplay (named a gamemode). This is intended to allow large mods that change large amounts of game logic (such as a randomizer) to have explicit control over how the game will function at certain points.

The selection for gamemodes is currently done on a button added to the pre-launch menu. The behavior of the vanilla pre-launch menu isn't changed by this PR (it only shows up at launch), but when there is a second gamemode registered, it will re-appear when resetting from the menu bar.

When registering a gamemode, a developer specifies a save file name to load. This will allow saves to be specific per-gamemode. To make this work I made some changes to the memcard manager class that allows the file name to be changed.

As a demonstration and to simplify the existing code, the speedrun mode was changed to be a gamemode that gets registered as the setting is enabled. The speedrun gamemode uses its own set of savefiles, different than the normal dusklight ones.


TakaRikka (commented):
kinda feel like there should be another more intuitive way of reaching the pre-launch besides resetting, otherwise people might not understand. maybe a button in settings or something

src/dusk/ui/modal.cpp
```diff
@@ -25,6 +25,10 @@ Modal::Modal(Props props) : WindowSmall("modal", "modal-dialog"), mProps(std::mo
 
     auto* actions = append(mDialog, "div");
     actions->SetClass("modal-actions", true);
+    if (props.isVertical) {
+        actions->SetClass("modal-actions-vertical", true);
+    }else{
+    }
```
encounter:
```suggestion
    }
```

src/dusk/ui/prelaunch.cpp
```diff
@@ -685,17 +687,85 @@ void try_apply_mirrored_layout(Rml::Element* body) {
...
+                menuList->RemoveChild(menuList->GetChild(0));
+            }
+            prelaunch->mMenuButtons.clear();
+            prelaunch->build_menu_buttons();
+            break;
+        }
+    }
+}
+
+static std::string getPlayButtonText() {
```
encounter: snake_case for internal functions

src/dusk/gamemode.hpp
```diff
@@ -0,0 +1,69 @@
...
+    std::string mFullName;
+    std::string mSaveName;
+
+    std::function<void()> mOnActivatedFunction = gamemodeStub;
+    std::function<void()> mOnDeactivatedFunction = gamemodeStub;
+    std::function<void()> mOnPlayFunction = gamemodeStub;
+    std::function<void()> mOnSaveLoadedFunction = gamemodeStub;
+    std::function<void()> mOnNewSaveFunction = gamemodeStub;
+    std::function<void()> mOnGameResetFunction = gamemodeStub;
+    std::function<void()> mOnTickFunction = gamemodeStub;
```
encounter: I'd just leave these unset and check them before calling

jdflyer:
@TakaRikka @encounter I made a few of the requested changes. I ended up adding a "Restart To Main Menu" button that appears under the "Open Data Folder" button within the settings, which I hope will be clear to the user. The logic still has the menu appearing from a menu-bar reset done while gamemodes are registered. Let me know if you guys have any other suggestions or thoughts about this feature.

TakaRikka:
think this looks fine to me (once the conflicts are fixed), but i defer to enc

TakaRikka (approved):

src/dusk/mods/svc/gamemode.cpp
```diff
@@ -0,0 +1,153 @@
...
+};
+
+}
+
+constinit const ServiceModule g_gamemodeModule{
+    .id = GAMEMODE_SERVICE_ID,
+    .majorVersion = GAMEMODE_SERVICE_MAJOR,
+    .minorVersion = GAMEMODE_SERVICE_MINOR,
+    .service = &s_gamemodeService,
+    .modDetached = gamemode_impl::gamemode_remove_mod,
```
encounter:
```suggestion
    .modDeactivating = gamemode_impl::gamemode_remove_mod,
```
modDetached runs _after_ the mod has shutdown, so I added `modDeactivating` for things that should happen _before_ shutdown (this can call into `unregisterGamemode`)

src/m_Do/m_Do_main.cpp
```diff
@@ -502,6 +502,56 @@ static void log_build_info() {
     DuskLog.info("Platform: {}", BOREALIS_PLATFORM_NAME);
 }
 
+static void mods_init(const std::string& mods_dir) {
```
encounter:
```suggestion
static void mods_init(const std::filesystem::path& mods_dir) {
```

encounter (approved):

PR #2231, opened 2026-07-19
Fixed connection issues, item dupe issues and certain flags not being set

hellkith:
## Summary

Fixed the following issues:
- crashing when retrieving settings from the  AP server causing it to  use  the default settings.
- item duping (small keys, sky characters, poe souls etc.)
- items not being retrieven during a big batch  (async AP run, more than 10 items received since last play)
- Certain flags not being set (goron elders in  the mines, Iza minigame when you don't teleport up there)
- Preventing items beinig send/obtained twice (ie progressive sword on heart container, poe soul on Iza helping hand)
- Added a timeout to the connection on the save file menu. If AP doesn't respond, you don't get stuck
- Save file losing AP link if autosave overwrites it with an empty string due  to connection issues or accidentally cancelling the connection on the save file screen.

I've run through it several times in a solo AP with DS3 as my second game.. Last 2 runs went flawless.


PJB3005:
Uh, development of this PR probably should work against the modern implementation of rando, which is [As a mod](https://github.com/TwilitRealm/dusklight-randomizer/)

PR #2237, opened 2026-07-24
Audio Fix for THP Movie Playback + Lock-Free Message Queue

iAmSomeone2:
## tldr;

This PR fixes THP movie audio crackling issues. It also adds a new Vyukov-style lock-free message queue system as a replacement for the original MessageQueue/SideTable implementations in 'src/dusk/stubs.cpp'.

## Background

A couple weeks back I happened to let the title screen play out all the way to the attract movie that plays after the logo pops up. Strangely, I was getting some pretty bad audio crackling as the movie played. On top of that, the issue was identical on both my M4 Pro MacBook and Ryzen 5800X-based Linux PC, so I knew it wasn't environment-specific. I ended up digging through the code for quite a while for what ended up being a fairly simple fix.

The first thing that stuck out to me was how much lock contention was present in the existing MessageQueue system. Every 'push' or 'pop' from a queue would lock a global mutex, preventing any other thread from using the queue. With the movie playback code making constant use of these queues and flame charts showing a lot of wait time at those points, that's what made the most sense to tackle.

I implemented and tested new lock-free bounded queue and side table implementations compatible with the original GC engine code. The lock contention was gone, but the crackling remained. Since the new queue was a lot of work and did remove those lock contention issues, I chose to keep it in anyway.

Some more digging and Tracy plotting ended up showing me the real culprit: SDL's bursty audio pipeline. The original hardware's DSP consumes the audio buffer at a constant rate, so it was perfectly fine to only buffer ~100ms of audio at a time. SDL 3 doesn't do that, instead it pulls bursts of data at regular intervals. My measurements recorded SDL requesting as high as 165ms of audio in a single burst. When the buffers couldn't satisfy the burst, SDL would fill the gap with silence. Those 1-65ms gaps would then sound like crackling audio!

Fixing that was simple enough on the surface; just make the buffer bigger on the `TARGET_PC` builds. Unfortunately, the original buffers are allocated using `JKRHeap` meaning that there wasn't space to make the buffer bigger. I also couldn't use a plain vector allocated at runtime or even a static array. The THP decoding logic requires the buffer to be 32-bit aligned. Anything else and the engine crashes. `alignas` couldn't make that work either. To get that working, I ended up making `PCHeap`. It's a basic compile-time sized memory pool with a bump allocation scheme. It also features a bit of optional Tracy instrumentation.

## Known Issues

- There's a pre-existing race condition that can randomly occur at the end of movie playback which sometimes causes the game to sit at a black screen indefinitely. It's most likely caused by one of the movie player threads not waking up correctly.

## Why Did Claude Mark Itself as an Author?

All of my experimentation made a mess of my git history, and asking Claude to help me untangle it was much faster than trying to do it myself. I did not realize that it was going to also write the commit message and list itself as a maintainer 🤦‍♂️.

To be clear, that is the *only* part of this PR that Claude has touched. Otherwise, my usage consisted of the following:

- Root cause analysis assistance.
- Problem space research.
- Codebase structure analysis.
- Work plan development.
- Rubber ducky.
- Initial code review.

I can attest that all new code and comments are my own work.


SuperDude88:
A lot of these changes seem excessive/aren't changes we currently want to make

PR #2243, opened 2026-07-25
Start annotating functions with override properly

PJB3005:
This is to mute compiler warnings and to avoid future mistakes like what previously happened with JUTResFont

These are behind a macro, DUSK_OVERRIDE, to make it clear they are Dusk modifications to the decomp. This is similar to DUSK_CONST etc.

This doesn't annotate even close to all the overrides but it's a start.

### DIFF
--- a/include/global.h
+++ b/include/global.h
@@ -255,5 +255,6 @@ using std::isnan;
 
 #define DUSK_CONST IF_DUSK(const)
 #define DUSK_CONSTEXPR IF_DUSK(constexpr)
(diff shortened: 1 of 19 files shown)


encounter:
Could bring over my helpers that allow using `override` and `final` in matching code: https://github.com/PrimeDecomp/prime/blob/f1f179dca93c5d212c29c756082e73f76a25b4d5/extern/sdk/include/dolphin/types.h#L75-L96

PR #2257, opened 2026-07-31
Fix AIInitDMA stub signature

Irastris:
An Aurora submodule update broke building due to a change to AIInitDMA, this fixes that


Irastris:
Or not? It worked on Windows with CLion + MSVC but CI still fails globally, not sure why

Irastris:
Superseded by https://github.com/TwilitRealm/dusklight/pull/2256, which I just noticed also included a fix for this problem

PR #2269, opened 2026-08-04
Fix for Linux distributions that use lib64: force libturbojpeg to use 'lib'

Pwootage:
After switching to Fedora (and installing the relevant dependencies) I get the following compile error when building (linux-default-relwithdebinfo):

```/usr/bin/ld.bfd: cannot find libjpeg-turbo-install/lib/libturbojpeg.a: No such file or directory```

Looking at the build log, I see the following:

```-- Up-to-date: /home/pwootage/projects/dusklight/build/linux-default-relwithdebinfo/libjpeg-turbo-install/lib64/libturbojpeg.a```

So it appears there is some issue here with 'lib' vs 'lib64'.  This seems to be the simplest solution.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,6 +150,7 @@ if (DUSK_MOVIE_SUPPORT)
             -DENABLE_SHARED=OFF
             -DWITH_TURBOJPEG=ON
             -DWITH_JAVA=OFF
+            -DCMAKE_INSTALL_LIBDIR=lib
         )
         if (CMAKE_TOOLCHAIN_FILE)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
encounter, line 145: Maybe instead we use `${CMAKE_INSTALL_LIBDIR}` here?

Not sure if that's meaningfully better

Pwootage: I guess the biggest problem there would be if somehow turbojpeg's CMAKE_INSTALL_LIBDIR was different than Dusk's; testing it, it does seem to work. I don't know why that would ever be different, though.

Are there any problems with cross compiling? I can't imagine there would be, since LIBDIR would be set correctly

encounter: Thought about it more and this solution is good. Thanks!

PR #2270, opened 2026-08-05
Mod SDK: Actor service

jdflyer:
This service implements a way for mods to register a custom actor that will be run by the game's engine. These actors can be loaded either by custom stages or by an assigned process name given at registration time.


TakaRikka (commented):
think this looks good afaict. maybe would be good to have a sample mod for it though?

TakaRikka (approved):

PR #2280, opened 2026-08-06
Mod SDK: Arc Overlays

jdflyer:
This implements the ability to a mod to supply overlay files when an archive is used as a directory, removing the need for mod authors to publish full archive files with their mods.

Let me know if you have any questions or thoughts about the implementation, or any ways that could make it more simple.


libs/JSystem/src/JKernel/JKRMemArchive.cpp
```diff
@@ -149,6 +157,14 @@ void* JKRMemArchive::fetchResource(SDIFileEntry* fileEntry, u32* resourceSize) {
 void* JKRMemArchive::fetchResource(void* buffer, u32 bufferSize, SDIFileEntry* fileEntry,
                                    u32* resourceSize) {
     JUT_ASSERT(595, isMounted());
+
+#ifdef TARGET_PC
+    void* overlay_data = getOverlayData(fileEntry, resourceSize);
+    if (overlay_data) {
+        return overlay_data;
```
encounter: This function must write out to `buffer`, the return value appears to be ignored

libs/JSystem/src/JKernel/JKRArchivePri.cpp
```diff
@@ -270,4 +271,93 @@ void JKRArchive::initFileDataPointers() {
...
+    if (DVDConvertPathToEntrynum(mArcOverlaysPath.c_str()) < 0) {
+        return nullptr;
+    }
+
+    // Build a map of file Ids -> Local path names
+    if (mIdToPathMap.empty()) {
+        buildIdToPathMap(0, std::string(&mStringTable[mNodes[0].name_offset])+"/");
+    }
+
+    const auto& it = mFileIdToArcOverlayData.find(fileEntry->index);
```
encounter: This uses `fileEntry->index` but inserts using `fileEntry->file_id` below

libs/JSystem/include/JSystem/JKernel/JKRArchive.h
```diff
@@ -211,6 +215,23 @@ class JKRArchive : public JKRFileLoader {
 
 #if TARGET_PC
     void** mFileData;
+
+    std::unordered_map<s32, std::vector<u8>> mFileIdToArcOverlayData;  // Owns all overlayed data
```
encounter: I think we're going to have to invalidate this when overlays change, right?

libs/JSystem/src/JKernel/JKRArchivePri.cpp
```diff
@@ -270,4 +271,93 @@ void JKRArchive::initFileDataPointers() {
...
+    s32 status = DVDReadPrio(&fileInfo, buffer.data(), ALIGN_NEXT(fileInfo.length,0x20), 0, 2);
+    DVDClose(&fileInfo);
+
+    if (status < DVD_RESULT_GOOD) {
+        return nullptr;
+    }    
+
+    void* out_data = buffer.data();
+    mFileIdToArcOverlayData[fileEntry->file_id] = std::move(buffer);
+    return out_data;
```
encounter, line 362: The returned buffer here isn't registered with `JKAR_DATA` or anything, so calls like `getResSize` and similar will fail. Not yet sure what a good solution here will be

libs/JSystem/src/JKernel/JKRArchivePri.cpp
```diff
@@ -270,4 +271,93 @@ void JKRArchive::initFileDataPointers() {
...
+        return;
+    }
+
+    std::string path = std::string(buf);
+    size_t pos = path.rfind(".arc");
+
+    if (pos == std::string::npos) {
+        return;
+    }
+    path[pos] = '_'; // Replace the .arc with _arc
```
encounter: We decided that we will want to remove the `.arc` extension instead of replacing it with `_arc`

encounter (approved):

PR #2289, opened 2026-08-08
mods: add Android Keystore secret storage service

joeblack2k:
## Summary

- add a generic per-mod secret-storage service ABI
- use Android Keystore AES-256-GCM with authenticated, atomically persisted metadata
- isolate records by application ID, mod ID, and key
- wipe native and Java byte buffers across success, failure, free, and mod detach
- return `UNAVAILABLE` on platforms without a secure backend

## API

`SecretStorageService@1.0` provides `get`, `put`, `remove`, and owner-checked
`free`. Keys are limited to 128 bytes and values to 64 KiB. The host derives
the namespace from the calling `ModContext`; mods cannot choose another mod's
namespace.

## Validation

- focused C ABI and linked C++ core tests: passed
- Java crypto/state-machine assertion tests: passed
- Android source-contract scan: passed
- AddressSanitizer + UndefinedBehaviorSanitizer: passed
- ThreadSanitizer: passed
- official `android-arm64` CMake/Ninja preset: passed, including final `libmain.so`
- Gradle `:app:compileDebugJavaWithJavac`: passed
- `git diff --check`
- independent Luna Max review: PASS

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -297,6 +297,9 @@ set(DUSK_COPYRIGHT "Copyright (C) Twilit Realm contributors")
 
 source_group("dolzel" FILES ${DOLZEL_FILES} ${Z2AUDIOLIB_FILES} ${REL_FILES})
 source_group("dusklight" FILES ${DUSK_FILES})
+if (ANDROID)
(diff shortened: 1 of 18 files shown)


SuperDude88:
Not a service we currently plan to support/is a lot of code to maintain for a very specific function (+ doesn't seem like it has many clear use cases)

PR #2290, opened 2026-08-09
Add Aiming Reticle for Bow and Slingshot

### LINKED ISSUE #2186: Add an optional reticle/red dot to aim-able items (Slingshot, Bow/Arrow, etc).
### Feature Description

Please consider adding some form of reticle/red dot for aim-able items such as bow/arrow, slingshot, etc. that other 3D Zelda games had.  This might be a candidate for the cheat menu for this game.

### Platform

_No response_

### Additional Information

_No response_

baxlek:
Adds an aiming reticle to first-person for bow and slingshot. Utilizes code that was already present, just disabled.

Closes #2186.

### DIFF
--- a/src/d/actor/d_a_alink_bow.inc
+++ b/src/d/actor/d_a_alink_bow.inc
@@ -551,7 +551,7 @@ void daAlink_c::setBowSight() {
         getArrowFlyData(&dist, &speed, TRUE);
         checkSightLine(dist, &sight_pos);
         mSight.setPos(&sight_pos);
-        mSight.offDrawFlg();
+        mSight.onDrawFlg();
     } else {
         mSight.offDrawFlg();
     }
--- a/src/d/actor/d_a_alink_bow.inc
+++ b/src/d/actor/d_a_alink_bow.inc
@@ -543,18 +543,7 @@ void daAlink_c::setBowNormalAnime() {
 }
 
 void daAlink_c::setBowSight() {
-#if TARGET_PC
-    if (checkBowChargeWaitAnime() && !dComIfGp_checkPlayerStatus0(0, 0x200000) && dusk::getSettings().game.aimingReticle) {
-        cXyz sight_pos;
-        f32 dist;
-        f32 speed;
-
-        getArrowFlyData(&dist, &speed, TRUE);
-        checkSightLine(dist, &sight_pos);
-        mSight.setPos(&sight_pos);
-        mSight.onDrawFlg();
-#else
-    if (checkBowChargeWaitAnime() && !dComIfGp_checkPlayerStatus0(0, 0x200000)) {
+    if (checkBowChargeWaitAnime() && !dComIfGp_checkPlayerStatus0(0, 0x200000) IF_DUSK(&& dusk::getSettings().game.aimingReticle)) {
         cXyz sight_pos;
         f32 dist;
         f32 speed;
@@ -563,7 +552,7 @@ void daAlink_c::setBowSight() {
         checkSightLine(dist, &sight_pos);
         mSight.setPos(&sight_pos);
         mSight.offDrawFlg();
-#endif
+        IF_DUSK(mSight.onDrawFlg());
     } else {
         mSight.offDrawFlg();
     }
--- a/src/d/actor/d_a_alink_bow.inc
+++ b/src/d/actor/d_a_alink_bow.inc
@@ -553,16 +553,6 @@ void daAlink_c::setBowSight() {
         checkSightLine(dist, &sight_pos);
         mSight.setPos(&sight_pos);
         mSight.onDrawFlg();
-    }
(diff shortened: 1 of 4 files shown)


src/d/actor/d_a_alink_bow.inc
```diff
@@ -543,6 +543,17 @@ void daAlink_c::setBowNormalAnime() {
...
+        cXyz sight_pos;
+        f32 dist;
+        f32 speed;
+
+        getArrowFlyData(&dist, &speed, TRUE);
+        checkSightLine(dist, &sight_pos);
+        mSight.setPos(&sight_pos);
+        mSight.onDrawFlg();
+#else
     if (checkBowChargeWaitAnime() && !dComIfGp_checkPlayerStatus0(0, 0x200000)) {
```
Irastris, line 546: Do an `IF_DUSK(&& dusk::getSettings().game.aimingReticle` at the end of the original condition, instead of repeating the entire condition in its own TARGET_PC block

Irastris (changes requested):

src/d/actor/d_a_alink_bow.inc
baxlek: Done. I think. I've never used an IF_DUSK before. I got it built and working on my test branch.

src/d/actor/d_a_alink_bow.inc
```diff
@@ -543,7 +543,7 @@ void daAlink_c::setBowNormalAnime() {
 }
 
 void daAlink_c::setBowSight() {
-    if (checkBowChargeWaitAnime() && !dComIfGp_checkPlayerStatus0(0, 0x200000)) {
+    if (checkBowChargeWaitAnime() && !dComIfGp_checkPlayerStatus0(0, 0x200000) IF_DUSK(&& dusk::getSettings().game.aimingReticle)) {
```
MelonSpeedruns: Shouldn't this be a || instead of a && ?

Currently if I'm reading this correctly, disabling the Aiming Reticle option would disable reticles for everything, including the bow.

MelonSpeedruns (dismissed):

src/d/actor/d_a_alink_bow.inc
baxlek: It's only affecting the bow and slingshot because it's in d_a_alink_bow.inc. Each other file has their own reticle.

baxlek: I just built a test build using ||. Before, the reticle would only show while drawing an arrow or a seed. Similar to when readying the clawshot. Now, the reticle is active whenever the player is using the bow or slingshot in first-person. Regardless of if nocking an arrow/seed or not. Which implementation would be prefered?

Personally, I prefer as-is. Because it matches the behavior of the other aim-able items.

Disabling the Aiming Reticle option still shows the reticle when drawing an arrow/seed (not intended).

Irastris: Only when ammo is nocked, yes. I will be re-reviewing this soon.

Irastris: Drop the IF_DUSK here (sorry)

src/d/actor/d_a_alink_bow.inc
```diff
@@ -552,6 +552,7 @@ void daAlink_c::setBowSight() {
         checkSightLine(dist, &sight_pos);
         mSight.setPos(&sight_pos);
         mSight.offDrawFlg();
+        IF_DUSK(mSight.onDrawFlg());
```
Irastris: Change these two lines to just `DUSK_IF_ELSE(dusk::getSettings().game.aimingReticle ? mSight.onDrawFlg() : mSight.offDrawFlg(), mSight.offDrawFlg());` so that we aren't being redundant

Irastris (changes requested):
Alright, got a minute to look at this again and realized a better solution. Sorry for misguiding you with the `IF_DUSK`, this newly requested solution is even less intrusive and I will merge the PR once implemented

src/d/actor/d_a_alink_bow.inc
baxlek: Done, built, tested and working. Like I said, I've never used a IF_DUSK or DUSK_IF_ELSE so I'm deferring to your expertise

Irastris: The DUSK_IF_ELSE should replace the original `mSight.offDrawFlg()`, not be below it

baxlek: I always thought replacing stuff was a big no-no, that's why I assumed to leave it in. Sorry.

Irastris: That's the purpose of DUSK_IF_ELSE, it works for changing single lines to do your new logic when compiling with TARGET_PC and the original logic otherwise, same as if you'd used TARGET_PC ifdefs👍

Irastris (approved):

PR #2291, opened 2026-08-09
Add Built-In Cosmetics Mod

gymnast86:
Ports over the cosmetics options from the `randomizer` branch into a standalone mod. 

Supports recoloring the following:
- Hero's Tunic
- Zora Armor
- Wooden/Ordon/Master Sword
- Lantern Light
- Gale Boomerang
- Iron Boots
- Spinner
- A/B/X/Y/Z button UI
- Hearts (UI, heart drops, pieces of heart, and heart containers)
- Midna's Hair
- Midna's Charge Ring
- Link's Hair
- Wolf Link
- Epona

Also moves declaration of `PaneCache` arrays into source files instead of header files because they were causing naming conflicts.


mods/basic_cosmetics_mod/src/color_utils.hpp
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+/**
+ * File originally copied from console TPR with permission from isaac
+ * https://github.com/zsrtp/libtp_rel/blob/master/include/util/color_utils.h
+ */
+
+#include "dolphin/gx/GXStruct.h"
```
TakaRikka: probably should just be `#include <gx.h>`

mods/basic_cosmetics_mod/src/option_descriptions.hpp
```diff
@@ -0,0 +1,354 @@
+#pragma once
+
+inline const char* kLinksCapColorsExplanation =
```
TakaRikka: better as constexpr?

mods/basic_cosmetics_mod/src/midna_hair_color.hpp
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#include "dolphin/gx/GXStruct.h"
```
TakaRikka: `<gx.h>`

TakaRikka (approved):

mods/basic_cosmetics_mod/src/option_descriptions.hpp
GiorgioMendieta, line 1: Is it possible to extract the common colors into a string or const char* and just append it to every explanation instead of duplicating it every time ?

gymnast86: The common colors have a few different values depending on what cosmetic they're referring to. We could separate out all the unique parts and build/store them at runtime, but I felt like it was less work and less confusing to just have all the descriptions laid out in their entirety.

GiorgioMendieta: My bad I didn't realise there were different values, I thought it was all the same set of colors

mods/basic_cosmetics_mod/src/hooks.cpp
```diff
@@ -0,0 +1,694 @@
...
+#include "SSystem/SComponent/c_phase.h"
+
+#include <optional>
+
+// Main hook for recoloring textures on load
+DEFINE_HOOK_SYMBOL(
+    "mDoDvdThd_mountArchive_c::execute", s32(mDoDvdThd_mountArchive_c*), MountArchiveExecute);
+void mount_archive_execute_post(ModContext*, void* args, void* retval, void* userdata) {
+    auto archive = mods::arg<mDoDvdThd_mountArchive_c*>(args, 0);
+    handle_texture_overrides_on_load(archive);
```
encounter, line 48: We'll have to be very careful here: this executes on the DVD thread, whereas the mod may update state or even get unloaded on the main thread while this is executing. Is there a better place to hook this?

mods/basic_cosmetics_mod/src/midna_hair_color.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+#include <array>
+
+static constexpr std::array<std::array<GXColor, 9>, 10> midnaHairColors = {{
+    // Default
+    {{
+        /*l_lNormalKColor*/ /*l_normalKColor*/  /*l_normalColor*/
+        {0xFF, 0xDC, 0x00}, {0xB4, 0x87, 0x00}, {0x50, 0x00, 0x00},
+        /*l_bigKColor*/     /*l_lBigColor*/     /*l_bigColor*/
+        {0x50, 0x00, 0x00}, {0xFF, 0x78, 0x00}, {0xFF, 0x64, 0x78},
```
encounter, line 14:
```suggestion
        {0x1E, 0x00, 0x00}, {0xFF, 0x78, 0x00}, {0xFF, 0x64, 0x78},
```
from 
```cpp
static GXColor const l_bigKColor = { 0x1E, 0x00, 0x00, 0x00 };
```

mods/basic_cosmetics_mod/src/midna_hair_color.cpp
```diff
@@ -0,0 +1,141 @@
...
+        /*l_lNormalKColor2*//*l_normalKColor2*/ /*l_lBigKColor2*/
+        {0x00, 0xC3, 0xEB}, {0xC3, 0xC3, 0x00}, {0xAA, 0xFF, 0xC3}
```
encounter, line 16:
```suggestion
        {0x00, 0xC3, 0xEB}, {0x00, 0xC3, 0xC3}, {0xAA, 0xFF, 0xC3}
```
from
```cpp
static GXColor const l_normalKColor2 = { 0x00, 0xC3, 0xC3, 0x00 };
```

mods/basic_cosmetics_mod/src/texture_utils.cpp
```diff
@@ -0,0 +1,549 @@
...
+    dst[1] = static_cast<uint8_t>(c0_565 & 0xFF);
+    dst[2] = static_cast<uint8_t>(c1_565 >> 8);
+    dst[3] = static_cast<uint8_t>(c1_565 & 0xFF);
+    dst[4] = static_cast<uint8_t>(indices >> 24);
+    dst[5] = static_cast<uint8_t>((indices >> 16) & 0xFF);
+    dst[6] = static_cast<uint8_t>((indices >> 8) & 0xFF);
+    dst[7] = static_cast<uint8_t>(indices & 0xFF);
+}
+
+bool convert_i8_to_cmpr(ResTIMG* texHeaderPtr) {
```
encounter, line 286: This only converts the base mip level (not sure if it's guaranteed to be a single mip or not)

mods/basic_cosmetics_mod/src/mod.cpp
```diff
@@ -0,0 +1,282 @@
...
+    return MOD_OK;
+}
+
+MOD_EXPORT ModResult mod_update(ModError*) {
+    update_rainbow_rgb(1.0f);
+    set_all_midna_hair_colors();
+    return MOD_OK;
+}
+
+MOD_EXPORT ModResult mod_shutdown(ModError*) {
```
encounter, line 277: We should probably best-effort clean up resources here. The mod services will handle it if we don't, but this will be an example mod, after all

encounter:
Related to the DVD hook comment: we should build this around the `TextureService` ideally, so that replacements don't require actually changing the underlying data, and they get cleanly unloaded when the mod does. If TextureService needs new APIs, let me know

mods/basic_cosmetics_mod/src/texture_utils.cpp
gymnast86: All the textures we modify are single mip, but I went and added functionality to account for multiple mips.

PR #2292, opened 2026-08-09
initial item service implementation

TakaRikka:


encounter:
Because of changes to game structs, we should bump GameService version

encounter (approved):

PR #2293, opened 2026-08-10
request presentation sync for pause and map captures

matthewdavidrichardanderson:
expands on #1447 so when interpolation is on, the pause screen, map and item wheel all show the correct freeze frame, helpful for alternate back in time methods that dont use item wheel


encounter:
Looks good, thanks!

PR #2303, opened 2026-08-13
Use texture replacement API for HQ minimap textures

dooplecks:
Successor to #2260.

- Uses Aurora texture replacement API directly instead of mod system.
  - This gives us fine control over priority, allowing us to place them below user replacements.
- Toggled via config var, enabled by default, but not exposed in the UI.
  - Aligns with original implementation, which replaced texture data unconditionally.
  - The original textures are extremely low-res, and I don't think they're worth cluttering the settings with a option.


encounter:
I dig it

PR #2308, opened 2026-08-17
Add Additional "Always Greatspin" Option

baxlek:
Adds an option to the "Always Greatspin" cheat to allow the Great Spin attack at any health only after obtaining the skill.


TakaRikka (approved):

baxlek:
Not sure why, but whenever I activate Speedrun Mode, this is the only cheat that gets grayed-out in the vanilla gameplay mode.

Is the Speedrun Mode toggle under Gameplay settings being deprecated for a dedicated Speedrun gameplay mode à la Randomizer?

src/dusk/ui/settings.cpp
```diff
@@ -1300,8 +1306,40 @@ SettingsWindow::SettingsWindow(bool prelaunch) : mPrelaunch(prelaunch) {
...
-            "Allows the Great Spin attack without requiring full health.");
+        leftPane.register_control(
+            leftPane.add_select_button({
+                .key = "Always Greatspin",
+                .getValue =
+                    [] {
+                        return kAlwaysGreatspinModes[static_cast<u8>(
+                            getSettings().game.alwaysGreatspin.getValue())];
+                    },
+                .isDisabled = [] { return getSettings().game.speedrunMode.getValue(); },
```
TakaRikka: this should be changed to `{ return dusk::speedrun::isActive(); }` to fix the issue you were seeing

PR #2339, opened 2026-08-23
Unrestricted items

baxlek:
Adds a cheat that removes most restrictions on when and where items can be used. Needs testing. Could possibly cause soft-locks, crashing or any number of undesirable effects. Feedback welcome.


baxlek:
TODO:
[X] Cast Fishing Rod underwater
[X] Lantern use underwater (should behave like out of oil and not display meter)
[X] Set Horse Call to always "Nothing Happened..." when underwater
[X] Prevent item use while swimming
[X] Fix Iron Boots being immediately unequipped
[X] Fix first-person items in Castle Town

baxlek:
^^ That's the impression I got while I was working on this too. I had to change way too many things to get this to work properly. If the consensus is that this should be a mod, feel free to close this PR.

PJB3005:
Oops yeah, I was discussing this on Discord and added that tag without intending to add it to the PR yet.

Considering this feature feels like "spot the wacky crashes" I'm really not sure it makes much sense in the base game, yeah.

baxlek:
I haven't noticed any crashing yet. Using first-person items in Castle Town shows back-face culling though. And most items don't really do anything helpful. It is more of a "It's cool you can do that!" kind of thing.

Underwater B&C might be useful for rando. But rando is also a mod. 🤷

baxlek:
Mod available at https://github.com/baxlek/Unrestricted-Items/releases

PR #2348, opened 2026-08-25
Add "Hold Button to Mash" Options

### LINKED ISSUE #1781: Holding A instead of mashing A
I have wrist issues that make the goron wrestling almost impossible, could there be a feature to make it so you hold the A button instead of mashing it?

baxlek:
Adds an accessibility option to "Hold" a button press instead of having to rapidly press it (mashing).

Instances covered are:
-Sumo Wrestling
-Swimming while wearing the Zora Armor and surfacing while not
-After getting stunned by the Twilit Bloat (Mini-boss holding the final Lanayru Tear of Light)
-"Chance" clash with Ganondorf
-Clinging to and biting an enemy as Wolf Link
-Spinner activators

Closes #1781


TakaRikka (approved):

src/dusk/settings.h
```diff
@@ -182,6 +182,12 @@ struct UserSettings {
         ConfigVar<bool> buttonFishing;
         ConfigVar<bool> instantSaves;
         ConfigVar<bool> instantText;
+        struct {
+            ConfigVar<bool> sumo;
+            ConfigVar<bool> swimming;
+            ConfigVar<bool> twilitBloat;
+            ConfigVar<bool> ganondorf;
+        } holdToMash;
```
TakaRikka: forgot to mention, i think it'd actually be better simplifying this down to a single setting. im not sure that having separate settings for this is really worth it

baxlek: I got the impression from MelonSpeedruns that having more options was always better. I guess for easier maintainability going forward, that'd be a good idea. Just in case there's more instances that need to be added.

src/d/actor/d_a_alink_sumou.inc
```diff
@@ -185,6 +185,7 @@ int daAlink_c::procSumouMove() {
         return 1;
     }
 
+    IF_DUSK(const bool easySumo = DUSK_IF_ELSE(dusk::getSettings().game.holdToMash.sumo, false));
```
baxlek: Is it alright to leave these?

src/d/actor/d_a_alink_sumou.inc
```diff
@@ -200,7 +201,7 @@ int daAlink_c::procSumouMove() {
 
         sp8 = frameCtrl_p->getRate();
         if (mProcVar0.field_0x3008 != 0) {
-            if (doTrigger()) {
+            if (easySumo ? doButton() : doTrigger()) {
```
baxlek: Or should the DUSK_IF_ELSE just be used here?

src/dusk/speedrun.cpp
```diff
@@ -56,6 +56,10 @@ void resetForSpeedrunMode() {
     getSettings().game.damageMultiplier.setSpeedrunValue(1);
     getSettings().game.instantDeath.setSpeedrunValue(false);
     getSettings().game.noHeartDrops.setSpeedrunValue(false);
+    getSettings().game.holdToMash.sumo.setSpeedrunValue(false);
```
baxlek: Also, should or shouldn't be allowed in Speedrun Mode? Since a lot of QoL stuff is allowed.

TakaRikka: for now lets leave it disabled, but we'll have a discussion in the speedrunning community about how to handle it at a later time

TakaRikka (approved):

src/d/actor/d_a_spinner.cpp
```diff
@@ -753,8 +753,9 @@ int daSpinner_c::execute() {
         } else {
             current.pos = field_0xaa4;
 
+            IF_DUSK(const bool holdToMash = DUSK_IF_ELSE(dusk::getSettings().game.holdToMash, false);)
             if (mSpinnerTag != TAG_END) {
-                if (!dComIfGp_event_runCheck() && mTrigJump) {
+                if (!dComIfGp_event_runCheck() && (holdToMash ? mDoCPd_c::getHoldA(PAD_1) : mTrigJump)) {
```
baxlek, line 758: Since there's a different mTrigJump config for Wii, does this also need a separate '#if PLATFORM_WII'?

baxlek: I'm guessing not. Since the control scheme is currently the same for both on Dusklight.

TakaRikka:
conflict needs to be resolved

PR #2350, opened 2026-08-26
Apple TV (tvOS) support

ThaLuffy:
# Apple TV (tvOS) support

Adds tvOS as a build target: the app runs on an Apple TV, keeps its saves across
tvOS reclaiming storage, and can be given a disc image from a browser without a
Mac.

The change is almost entirely additive — 72 files, +6052/−5 — and nothing
outside `platforms/tvos/`, `scripts/tvos/` and the tvOS-gated blocks in
`CMakeLists.txt` affects other platforms.

## Why tvOS needs more than a new target

Two platform constraints drive most of the design:

**There is no durable storage.** tvOS gives an app ~500 KB of persistent local
storage (`NSUserDefaults`); everything else lives in purgeable `Library/Caches`
and can be reclaimed while the app is not running. A memory-card file left there
disappears without warning.

**There is no file dialog.** `open_file_dialog`'s document-picker path is
guarded `!TARGET_OS_TV` upstream, so a user has no way to hand the app a disc
image. Until now that meant baking one into the bundle at build time, which
makes a build personal to whoever compiled it.

## What is in here

**Build and packaging** — `tvos-default` preset support, `scripts/tvos/*` for
setup, provisioning, signing, install and launch, and an `xcassets` catalog
compiled during the build so every path produces one (see "CI" below).

**Save persistence** — the card file and small config files are mirrored into
`NSUserDefaults` on every write, and restored when they are missing. Writes are
triggered by save completion with a short debounce, the envelope is versioned
with per-entry digests, and a restore never overwrites a save that is present.

**Disc discovery** — scans the bundle, then `<data dir>/discs/` and the data
root. One candidate is verified automatically, several present a chooser.

**Network disc transfer** — with no disc anywhere, the app shows a URL. Opening
it on a phone or laptop gives an uploader that sends the image in resumable
8 MiB chunks. Bytes stage in `discs/.incoming/` and are renamed into place only
after `dusk::iso::validate` accepts them, so a partial upload is never offered
as playable.

## Design notes a reviewer may want

**The transfer subsystem is split for testability.** `http_parse.cpp` and
`upload_core.cpp` are pure — no sockets, no filesystem, no clock — and covered
by host tests. `server.cpp` owns the sockets and takes its validator by
injection, so the transport can be exercised without disc parsing.

**Resume follows the server, not the client.** The uploader advances by the
offset the server acknowledges and treats `409` as "re-sync and continue".
`judge_chunk` requires `offset == received`, which turns a retried chunk whose
response was lost into a refusal rather than a silent double append — the
failure mode that would otherwise corrupt an image with no symptom until
validation.

**Browser-side validation is only a guard.** The page reads the game id from a
raw `.iso` header and rejects a wrong game before uploading. It cannot do this
for compressed containers, because `borealis::disc` hashes the *decoded* disc —
which is also why one catalog hash covers every container format. The device's
own `dusk::iso::validate` is what authorises publishing; the accepted ids the
page uses are derived from the same table, so a disc the app accepts can never
be one the page rejects.

**Everything network-facing is tvOS-only.** `server.cpp` uses POSIX sockets, so
the transfer sources and the onboarding screen are in a tvOS-gated
`target_sources` block rather than the shared list.

**No new dependencies.** The HTTP server is hand-rolled and deliberately
minimal — two methods, no keep-alive negotiation, no chunked request bodies —
because the alternative was vendoring a general-purpose library for four
endpoints.

## CI

The workflow already had `x-tvos-ci` and a Rust tvOS toolchain step; only the
matrix entry was commented out. This enables it, and the shared upload step
produces a `tvos-appleclang-arm64` artifact in the same shape as iOS.

One change was needed to make that work: the icon catalog is now compiled by a
CMake custom command rather than by `scripts/tvos/build.sh`. CI runs
`cmake --build` directly, so the script-based approach would have produced a
bundle with no `Assets.car` — and since `Info.plist` names catalog sets that
only resolve inside a compiled catalog, the app would have shipped with no icon
and a device can refuse to install it.

## Testing

- **Host tests:** `tests/transfer/` (HTTP parsing, upload rules, and an
  end-to-end socket test covering chunked upload, replay refusal, resume, and
  byte-identical publish), plus `tests/save_mirror/` and
  `tests/disc_discovery/`.
- **CI:** all jobs green, including the four Apple matrix entries.
- **On device** (Apple TV 4K 3rd gen, tvOS 26.5): save restored byte-identical
  after the data container was destroyed and only the `NSUserDefaults` mirror
  put back; a 902 MB disc uploaded from a laptop, verified, published, and
  picked up by discovery without a relaunch.

## Known limitations

- **Installing needs a Mac.** AltStore does not support tvOS, so sideloading is
  via Xcode or `devicectl`. The artifact is therefore less directly usable than
  the iOS one.
- **An uploaded disc lives in purgeable storage** and can be reclaimed; a
  bundled one cannot. Re-supplying is a resumable upload rather than a rebuild,
  but it is not free.


SuperDude88:
> Note that we do not accept contributions that are primarily AI-generated

ThaLuffy:
@SuperDude88 Too bad. I get why you need a rule here, and that enforcing it means closing stuff you can't be sure about.

But this wasn't "point a model at the repo, paste the diff". I spent days on it, and most of that was just understanding tvOS. You get roughly 500KB of storage that survives. Everything else can be wiped whenever the system feels like it. There's no file picker either, so a user has literally no way to hand the app a disc image. Both of those had to be designed around. That's what the save mirror and the upload server are for.

It's tested properly, too. I wiped the entire data container and the save came back byte for byte from the mirror. I uploaded a 902MB disc from my laptop to the Apple TV over the network and played it. All nine CI jobs pass. Turning tvOS on in the workflow meant uncommenting a block someone had already put there.

I used AI heavily. Not going to pretend otherwise, and honestly most code being written today has AI in it somewhere, including plenty that lands in repos without anyone mentioning it. The decisions here were mine and I can explain any part of it.

So which rule is it? No AI-generated code at all, or no code the author can't stand behind? If it's the second one, ask me anything you like. If it's the first, fair enough, it's your project. I'll keep it as a fork.

Branch is at `ThaLuffy/dusklight:tvos` if it's useful to anyone.

PR #2357, opened 2026-08-29
Randomizer for linux: forces the no inline

Noaseto:
For linux builds I was prompted theses errors (see image).

Some symbols are inlined and the usage of the `DUSK_NOINLINE` makes sure they do not. I added the keyword before every function that is hooked because from my testing, some symbols could be inlined just a few times resulting sometimes in the vanilla behaviour with no error logs.

I also added the mangled version of 2 calls that the log warned me about, even though they are not needed for the windows version.

And lastly I had a crash due to the std::upper so I used the same bypass as aurora does for this.

I don't feel like its the cleanest way to manage the inline, I believe it could/should be done directly in the CMakeLists file, but it does compile and I managed to generate a seed and launch the game.

image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -428,7 +428,8 @@ if (CMAKE_CXX_COMPILER_FRONTEND_VARIANT STREQUAL "GNU")
                 $<$<COMPILE_LANGUAGE:CXX>:-fno-ipa-cp>
                 $<$<COMPILE_LANGUAGE:CXX>:-fno-ipa-cp-clone>
                 $<$<COMPILE_LANGUAGE:CXX>:-fno-ipa-sra>
-                $<$<COMPILE_LANGUAGE:CXX>:-fno-partial-inlining>)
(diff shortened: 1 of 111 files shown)


TakaRikka:
if a solution with compiler flags could be found instead it'd be preferred, but if not then this is probably acceptable

Noaseto:
It seems that some stuff can be edited too in the hook.cpp
```C
DEFINE_HOOK_SYMBOL("daDitem_c::execute", int(daDitem_c*), daDitem_c__execute);
// can become
DEFINE_HOOK(&daDitem_c::execute, daDitem_c__execute);
```

encounter:
Thanks for the PR! This was actually a deeper and more convoluted issue which I fixed in https://github.com/TwilitRealm/dusklight/commit/905da94b4013cd5850bd0e11cc9ac7cfdaf92f58. Turns out the existing compilation flags _were_ actually sufficient, they just weren't being applied.

PR #2363, opened 2026-08-31
Instant Death Fixes

### LINKED ISSUE #1552: Link cannot die of fall damage when instant death mode is activated
Link’s health _increases_ rather than decreasing to zero.

### LINKED ISSUE #1563: King Bulblin cannot kill Link when in instant death mode.
Tested on the first fight, just before Arbiter’s Grounds. His attacks seem to increase Link’s health.

SuperDude88:
Alternate take on #1899

Does not check that `ItemLifeCount` isn't NaN/inf/some other float value that can't be converted to an int properly. In theory the value could still be too big even for `int` so if desired I could add a clamp and/or NaN checks, though I don't think any cases currently hit that in practice.

The change to `d_a_alink_damage` is entirely optional (though in theory lets the base damage go higher without overflowing in some cases), open to removing it if it's deemed unnecessary

Fixes #1563 
Seems to fix #1552, though I am not sure exactly what scenario they used to cause it

### DIFF
--- a/src/d/d_meter2.cpp
+++ b/src/d/d_meter2.cpp
@@ -606,7 +606,12 @@ void dMeter2_c::moveLife() {
             life_count = (dComIfGs_getMaxLife() / 5) * 4;
         }
 
+#if TARGET_PC
+        int new_life = dComIfGs_getLife() + dComIfGp_getItemLifeCount(); // prevent an overflow with really large damage values
+#else
         s16 new_life = dComIfGs_getLife() + dComIfGp_getItemLifeCount();
+#endif
+
         if (new_life > life_count) {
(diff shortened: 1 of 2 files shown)


src/d/d_meter2.cpp
```diff
@@ -606,7 +606,12 @@ void dMeter2_c::moveLife() {
             life_count = (dComIfGs_getMaxLife() / 5) * 4;
         }
 
+#if TARGET_PC
```
Irastris: This could be a one-liner `DUSK_IF_ELSE(int, s16)`

Irastris (approved):

PR #2377, opened 2026-09-03
Mods: Support Luau mods

encounter:
This introduces the concept of mod "runtimes". The first mod runtime is a bundled `luau_runtime`. When a mod sets `"runtime": "dev.twilitrealm.luau@1.0"` in `mod.json`, the Luau runtime will load `res/main.luau` and execute it. The runtime itself being a (bundled) mod allows it to be updated independently of Dusklight, adding new modules and/or fixing bugs.

The runtime provides the following core set of modules to start:
- `dusklight.log` (`LogService`)
- `dusklight.host` (`HostService`)
- `dusklight.config` (`ConfigService`)
- `dusklight.resources` (`ResourceService`)
- `dusklight.overlay` (`OverlayService`)
- `dusklight.texture` (`TextureService`)
- `dusklight.ui` (`UiService`)

These allow Luau mods to, for example:
- Create interactive UIs (mod panel, windows, dialogs, etc.)
- Read/write config vars
- Dynamically enable/disable/swap out DVD overlays and texture replacements

Without having to ship a full native mod compiled for all platforms. Key constraint: Luau mods (currently) have no access to game code or hooks. Native C++ mods are still the primary way to develop full-featured mods that utilize all mod API features.

I also rewrote a large portion of `docs/modding.md` and classified mods into three categories (asset-only mods, native mods, and script mods).

### DIFF
--- a/docs/modding.md
+++ b/docs/modding.md
@@ -172,7 +172,7 @@ Available features:
 - `game`: Allows calling into and hooking game code. Mods that **only** use services may omit it, providing a wider
   range of compatibility with Dusklight versions and a slightly faster build process.
 - `webgpu`: Allows importing the WebGPU API
-  ([`webgpu/webgpu.h`](https://github.com/encounter/dawn/blob/main/include/webgpu/webgpu.h)). Must be enabled when using
+  ([`webgpu/webgpu.h`](https://github.com/webgpu-native/webgpu-headers/blob/main/webgpu.h)). Must be enabled when using
   [GfxService](#gfxservice-modssvcgfxh).
 
 Building produces `my_mod.dusk` in `build/mods/`. Copy the `.dusk` into the user mods folder:
@@ -944,7 +944,7 @@ New windows are hidden by default so a mod can finish attaching graphics before 
 **Requires `add_mod(... FEATURES webgpu)`**
 
 Direct WebGPU access at various stages of the rendering pipeline. Mods use the `wgpu*` C API (via
-[`webgpu/webgpu.h`](https://github.com/encounter/dawn/blob/main/include/webgpu/webgpu.h)) for custom draws and compute
+[`webgpu/webgpu.h`](https://github.com/webgpu-native/webgpu-headers/blob/main/webgpu.h)) for custom draws and compute
 dispatches. Mods must manage their own WebGPU state, including pipelines and bind groups.
 
 ```cpp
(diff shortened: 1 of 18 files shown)


TakaRikka (approved):

docs/modding.md
```diff
@@ -81,99 +183,130 @@ Passing `--mods <dir>` on the command line replaces the user directory with one
...
-All three lifecycle exports are required. `mod_ctx` is your mod's identity token, set by the loader before
-`mod_initialize` runs. Pass it as the first argument to every service call.
+To create a script mod, copy `sdk/luau/template`, edit its manifest and source, then zip the `mod.json` and adjacent
+folders and rename to `.dusk`. Copy the `.dusk` into the user mods folder:
+
+- Windows: `%APPDATA%\TwilitRealm\Dusklight\mods`
+- Linux: `~/.local/share/TwilitRealm/Dusklight/mods`
+- macOS: `~/Library/Application Support/TwilitRealm/Dusklight/mods`
+
+**Restrictions:** The Luau VM has no raw filesystem, network, or game-code access. Each script mod has a 64 MiB memory
```
PJB3005, line 273: Going over 250ms update limit unconditionally disables the mod? I'm not sure this is a good idea as e.g. OS page swapping stalls are unavoidable and not necessarily the mod's fault.

PJB3005 (changes requested):
Image

Doesn't compile in debug for me

mods/luau_runtime/src/runtime.cpp
```diff
@@ -0,0 +1,611 @@
...
+    } else {
+        const char* currentPath = lua_tostring(state, lua_upvalueindex(2));
+        const auto normalized =
+            normalize_module_path(currentPath != nullptr ? currentPath : "main.luau", requested);
+        if (!normalized.has_value()) {
+            luaL_error(
+                state, "module paths must be relative and remain inside the mod's res directory");
+        }
+        moduleName = *normalized;
+    }
```
PJB3005, line 278: Do we not need a system for other mods to register Lua modules?

PJB3005:
Actually the build failure is with Rando too, unrelated to this PR.

mods/luau_runtime/src/runtime.cpp
encounter: I wasn't going to include that yet, because it means we'd have to handle mod load ordering between Luau mods, pretty complicated

docs/modding.md
encounter: Fair, could relax this

PR #2379, opened 2026-09-03
Fix mod CRT linkage on MSVC debug

PJB3005:
CRT linkage for mods was being hardcoded to MultiThreadedDLL. This was causing the Luau PR and similar to not compile on debug due to mismatching CRT.

Also needs a change in Rando to remove another hardcoding.


encounter:
The original intention was to make mods compile against the release CRT in all configs, so that there is no debug vs release mod incompatibility. What error are you hitting exactly?

PJB3005:
> What error are you hitting exactly?

Sub libraries of the Luau runtime are still being compiled with debug CRT/iterator config. So there's a linker error:

clion64_SCZkiXNOoR

> The original intention was to make mods compile against the release CRT in all configs, so that there is no debug vs release mod incompatibility. 

This isn't really possible in principle unless we also make Dusklight itself always compiled against non-debug CRT, even on debug builds.

encounter:
> This isn't really possible in principle unless we also make Dusklight itself always compiled against non-debug CRT, even on debug builds.

That's not quite true, we don't share anything CRT-related between the host and mod, so there's nothing to conflict. But I think that means we should just _remove_ the overrides altogether and let the defaults win; we don't have to try to match the host after all.

PJB3005:
> That's not quite true, we don't share anything CRT-related between the host and mod

Strictly speaking, we do if the mod is interfacing with the game's C++ code/ABI directly. But yes, the overrides should just be removed.

encounter:
Removed in https://github.com/TwilitRealm/dusklight/commit/bc32702cb45a13b237aa01e28fd8b5fd22bd3f87

PR #2384, opened 2026-09-03
Basic discrete surround sound support

dooplecks:
- Implements discrete 5.1 and 7.1 audio output using a pairwise panner.
- Unifies regular stereo, spatial audio, and discrete surround output modes under a single setting in the Dusklight UI.
- Removes the audio output mode setting from the in-game UI.
- Original stereo output modes (both regular and spatial) remain functionally untouched.

Reverb handling is very rudimentary. Since Freeverb only operates in stereo, we downmix surround voices before adding them to the reverb input bus. After doing reverb processing, we upmix the ouput and mix it into the surround output bus. There is certainly room for improvement here, but it will probably involve a lot of experimentation and tuning. This work may come as part of a spatial audio rework.


TakaRikka (approved):

KoB-Kirito:
Awesome! Will be able to test tomorrow.

Edit: Works very well. Thank you <3

PR #2408, opened 2026-09-08
Separate music volume setting

kamilink7:
I don't know if this is the desired way to go about this, or if something in DuskAudioSystem would have been preferred, but this works as intended in testing. On another branch I just have this as a bool setting like with Recording Mode, so I can change this if that would be preferred.

### DIFF
--- a/src/f_ap/f_ap_game.cpp
+++ b/src/f_ap/f_ap_game.cpp
@@ -749,6 +749,7 @@ static void duskExecute() {
     dusk::menu_pointer::begin_game_frame();
     dusk::input::handleGamepadColor();
     updateAutoSave();
+    auto floatMusicVolume = static_cast<float>(dusk::getSettings().audio.mainMusicVolume.getValue()) / 100;
 
     if (dusk::getSettings().game.recordingMode) {
         Z2GetSoundMgr()->getSeqMgr()->getParams()->moveVolume(0.0f, 0);
@@ -760,7 +761,13 @@ static void duskExecute() {
         isRecording = false;
     }
 
-
+    if (floatMusicVolume != 1.0f) {
+        Z2GetSoundMgr()->getSeqMgr()->getParams()->moveVolume(floatMusicVolume, 0);
+        Z2GetSoundMgr()->getStreamMgr()->getParams()->moveVolume(floatMusicVolume, 0);
+    } else {
+        Z2GetSoundMgr()->getSeqMgr()->getParams()->moveVolume(1.0f, 0);
+        Z2GetSoundMgr()->getStreamMgr()->getParams()->moveVolume(1.0f, 0);
+    }
 
     if (dusk::getSettings().game.fastSpinner && mDoCPd_c::getHoldR(PAD_1)) {
         if (const auto link = g_dComIfG_gameInfo.play.getPlayer(0)) {
--- a/src/f_ap/f_ap_game.cpp
+++ b/src/f_ap/f_ap_game.cpp
@@ -751,6 +751,14 @@ static void duskExecute() {
     updateAutoSave();
     auto floatMusicVolume = static_cast<float>(dusk::getSettings().audio.mainMusicVolume.getValue()) / 100;
 
+    if (floatMusicVolume != 1.0f) {
+        Z2GetSoundMgr()->getSeqMgr()->getParams()->moveVolume(floatMusicVolume, 0);
+        Z2GetSoundMgr()->getStreamMgr()->getParams()->moveVolume(floatMusicVolume, 0);
+    } else {
+        Z2GetSoundMgr()->getSeqMgr()->getParams()->moveVolume(1.0f, 0);
+        Z2GetSoundMgr()->getStreamMgr()->getParams()->moveVolume(1.0f, 0);
+    }
+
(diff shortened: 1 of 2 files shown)


PJB3005 (dismissed):
I'm pretty sure the game already manually moves these parameters in some locations, which you're clobbering. You'll need to take those spots into account.

PJB3005 (approved):

src/f_ap/f_ap_game.cpp
```diff
@@ -760,7 +761,13 @@ static void duskExecute() {
...
     }
 
-
+    if (floatMusicVolume != 1.0f) {
+        Z2GetSoundMgr()->getSeqMgr()->getParams()->moveVolume(floatMusicVolume, 0);
+        Z2GetSoundMgr()->getStreamMgr()->getParams()->moveVolume(floatMusicVolume, 0);
+    } else {
+        Z2GetSoundMgr()->getSeqMgr()->getParams()->moveVolume(1.0f, 0);
+        Z2GetSoundMgr()->getStreamMgr()->getParams()->moveVolume(1.0f, 0);
+    }
```
PJB3005, line 770: Shouldn't the new logic be placed *before* recording mode? Or otherwise at least placed in such a way that recording mode overrides this?

PJB3005 (changes requested):

src/f_ap/f_ap_game.cpp
kamilink7: Oh good point, I'll see how they interact when I get home from work later. I'm not sure if the bool isRecording would force the audio scale to 0 already every loop or not, otherwise I can move it to before recording mode logic or put the new logic under an if (!dusk::getSettings().game.recordingMode)

kamilink7: Alright, looks like just placing it before fixed it like you said.

src/f_ap/f_ap_game.cpp
```diff
@@ -749,6 +749,15 @@ static void duskExecute() {
...
     updateAutoSave();
+    auto floatMusicVolume = static_cast<float>(dusk::getSettings().audio.mainMusicVolume.getValue()) / 100;
+
+    if (floatMusicVolume != 1.0f) {
+        Z2GetSoundMgr()->getSeqMgr()->getParams()->moveVolume(floatMusicVolume, 0);
+        Z2GetSoundMgr()->getStreamMgr()->getParams()->moveVolume(floatMusicVolume, 0);
+    } else {
+        Z2GetSoundMgr()->getSeqMgr()->getParams()->moveVolume(1.0f, 0);
+        Z2GetSoundMgr()->getStreamMgr()->getParams()->moveVolume(1.0f, 0);
+    }
```
PJB3005: Could you make the recording mode logic just set `floatMusicVolume` instead so that you don't need to copy/paste this code?

PJB3005 (changes requested):

src/f_ap/f_ap_game.cpp
kamilink7: Done; kept isRecording just in case it was planned to be used for something else in the future, otherwise recording mode just sets floatMusicVolume to 0.0f now

src/f_ap/f_ap_game.cpp
```diff
@@ -749,19 +749,21 @@ static void duskExecute() {
...
+    isRecording = false;
     if (dusk::getSettings().game.recordingMode) {
-        Z2GetSoundMgr()->getSeqMgr()->getParams()->moveVolume(0.0f, 0);
-        Z2GetSoundMgr()->getStreamMgr()->getParams()->moveVolume(0.0f, 0);
+        floatMusicVolume = 0.0f;
         isRecording = true;
-    } else if (isRecording) {
+    }
+
+    if (dusk::getSettings().game.recordingMode || floatMusicVolume != 1.0f) {
```
PJB3005: Isn't this `dusk::getSettings().game.recordingMode` check redundant?

PJB3005 (changes requested):

hawrz:
Ha oops i'm sorry but i have been working on separate audio sliders for a while, and testing it locally before creating my PR, and only now I notice yours :(
there is overlap, but i think i've done things differently (in jaudio) and added other sliders as well (SFX, voices, etc)
dont know which is best though \o/

kamilink7:
> Ha oops i'm sorry but i have been working on separate audio sliders for a while, and testing it locally before creating my PR, and only now I notice yours :(
> 
> there is overlap, but i think i've done things differently (in jaudio) and added other sliders as well (SFX, voices, etc)
> 
> dont know which is best though \o/ 

Honestly I'd just go ahead and PR yours and see which one they want to use more; sounds like yours might be a fair bit more advanced so I won't feel bad at all if yours is merged instead

src/f_ap/f_ap_game.cpp
kamilink7: Oh true, floatMusicVolume would already be set to 0 and meet the if condition anyway, I'll fix that when I get home from work

kamilink7: Fixed!

PJB3005 (approved):

Issue #564, opened 2026-04-26
Eldin Bridge cutscene softlock if Main HUD is disabled.


MelonSpeedruns:
It won't open the map. Disabling Main HUD forces the map to open and you can resume the game.

TakaRikka:
would it be better to just turn the HUD alpha down rather than disable it that way it doesnt cause logical breaks?

MelonSpeedruns:
Yeah that would be better actually. I can work on it either tonight or tomorrow.

baxlek:
The solution to this issue might also fix the graphical issues in issue #2107.

tomlube:
Fixed at some point

PR #2459, opened 2026-09-17
Quick Transform with Minimal HUD Fix

### LINKED ISSUE #2458: Unable to Quick Transform with Minimal HUD
### Mods and Forks

- [x] The bug occurs on an **official release** of Dusklight, without any forks or third-party patches installed

### Dusklight Version

1.4.1

### Bug Description

Unable to use Quick Transform after loading a new scene with Minimal HUD enabled.

### Steps to Reproduce

1. Enable Minimal HUD & Quick Transform settings
2. Transition to a new scene
3. Attempt to use Quick Transform (R+Y)

### Expected Behavior

Link should be able to transition to or from wolf form

### Current Behavior

The "action failed" sound effect plays.

### OS / Architecture

Android ARM64

### Game Disc

GameCube NTSC-U (North America)

### Game Language

_No response_

### Graphics Backend

Vulkan

### Texture Packs and Mods

None N/A

### Logs and Other Files

_No response_

### Additional Context

Probably related to the Midna icon not being loaded. After toggling Minimal HUD off then on again, Quick Transform is able to be used.

baxlek:
Fixes not being able to use Quick Transform and Sun's Song after a scene transition while Minimal HUD setting is active.

Closes #2458


MelonSpeedruns:
This won't work, as enabling minimal HUD will bypass the zalpha check. I think the zalpha check needs to be changed altogether.

baxlek:
What does the zalpha check even prevent? Using Quick Transform while the Call Midna button is dimmed? I tried to activate Quick Transform while the button would normally be dimmed (like when swinging a sword) with minimal HUD active and didn't notice anything unusual.

The solution to #564 would probably be the proper solution.

SuperDude88:
The game normally dims the Z button when you're not supposed to be able to talk to Midna -> transform, so the only way to really be confident it won't have an edge case that breaks is so check the conditions that would dim the Z button

baxlek:
I kinda consider this whole issue to be an edge case. If if's preferable to leave it as-is, I can live with that.

When loading a new scene with Minimal HUD active, the alpha value, I'm guessing, would currently be null. Not zero. Since the HUD element isn't even loaded. I'm guessing.

baxlek:
@MelonSpeedruns I think you're getting `||` and `&&` confused again.

PR #2461, opened 2026-09-18
Add Super Clawshot Option

baxlek:
Adds a Super Clawshot option (in addition to the current one) that only enables the "Attach to Everything" part of the full enhancement.


baxlek:
Just had an idea to make both parts individually selectable. Let me cook.

baxlek:
Incorporated into the Unrestricted Items mod.

PR #2464, opened 2026-09-18
Remove initTexObj debug log spam

PJB3005:


MelonSpeedruns:
Maybe just comment it out? Maybe someone could have a use to it to fix a bug at some point.

GiorgioMendieta:
> Maybe just comment it out? Maybe someone could have a use to it to fix a bug at some point.

Why not put it behind LOG_DEBUG log level instead ?

PJB3005:
> Why not put it behind LOG_DEBUG log level instead ?

It already is.

PJB3005:
> Maybe just comment it out?

Sure

PR #2497, opened 2026-09-19
Add bundled gamecontrollerdb and fallback mappings

999sian:
Adds community gamecontrollerdb.txt to resources and registers fallback mappings for Mayflash/DragonRise GameCube adapters across platforms, while continuing to respect user overrides in dusk::ConfigPath.

This resolves controllers and adapters not being recognized out of the box when operating in PC/DInput mode.


Irastris:
We used to include this ourselves, it was removed because it just causes more trouble than it solves in our experience. Users are free to add gamecontrollerdb to their own installs if they need to, we will not be shipping it again.

Abzol:
First of all surely this shouldnt be vendored since its comes outright from another repo

Secondly we already added this once and it broke 70% of all controllers. Dusklight currently loads gamecontrollerdb if it exists on paths the game reads

999sian:
Thanks for the context and explanation! Completely understand avoiding the full community database given the regressions with Android and Switch Pro controllers.

Would you be open to a minimal change that only registers targeted fallback mappings in code strictly for the Mayflash / DragonRise GameCube adapter (`0079:1844` / `0079:1843`) without any vendored database file, or is project policy strictly to upstream any new mappings to SDL3?

encounter:
We'd prefer upstreamed SDL3 mappings!

999sian:
Understood, thanks Luke! That makes total sense to keep all mapping definitions in upstream SDL3.

PR #2515, opened 2026-09-19
Fix map glitch crash w/ interpolation

### LINKED ISSUE #2513: Unplugging controller and Map Glitch Inputs crashes
### Mods and Forks

- [x] The bug occurs on an **official release** of Dusklight, without any forks or third-party patches installed

### Dusklight Version

2.0.0

### Bug Description

If you have interpolation on, then in *some* areas, unplugging the controller and doing the Map Glitch inputs (Z + Dpad left/right) the game crashes with EXCEPTION_ACCESS_VIOLATION. Unplugging and replugging the controller lags the game for a bit which I don't think the interpolation doesn't account for when you open two menus. It happens for calling Midna and pulling the item wheel at the same time as well.

All we know is that the issue exists due to interpolation which probably didn't account for the lag caused by replugging the controller.

### Steps to Reproduce

Precondition: Set unlock framerate to something other than off

1. Go to an area like Bulblin Camp
2. Unplug the controller
3. Hold Dpad right and Z, then replug the controller

### Expected Behavior

This should result in opening the map and storing Midna for later to perform Map Glitch, a technique used commonly for speedruns. Unplug strat is the safest method and thus commonly used by speedrunners.

### Current Behavior

The Game crashes with EXCEPTION_ACCESS_VIOLATION

Unfortunately we didn't figure out why it happens only in some areas. For example it does not happen in a big area like Lake Hylia but it does in an important area like Bulblin Camp.

### OS / Architecture

Windows x86-64

### Game Disc

GameCube NTSC-U (North America)

### Game Language

_No response_

### Graphics Backend

Unsure / Default

### Texture Packs and Mods

_No response_

### Logs and Other Files

[dusklight-20260919-225910.log](https://github.com/user-attachments/files/32424312/dusklight-20260919-225910.log)

### Additional Context

_No response_

encounter:
Not super happy with the solution here but it appears to work.

The problem appears to be that multiple simulation ticks can run after the menu locks the UI heap but before the menu capture painter sets the pause flag. Midna's message init runs without any resources and crashes. The hacky fix is to just skip simulation until we run the next paint.

Fixes #2513


spicyjuice04:
Can confirm that the crash now is fixed, didn't encounter an issue with Map Glitch anymore

TakaRikka:
this'll do for now, may revisit in the future

Issue #2503, opened 2026-09-19
Music Glitch Opening Scene


alfredo93patryluk:
### Mods and Forks

- [x] The bug occurs on an **official release** of Dusklight, without any forks or third-party patches installed

### Dusklight Version

2.0.0

### Bug Description

I've been playing Dusklight on both PC and Android and version 2.0.0 has introduced a new audio bug in the opening scene. This is present in both Android and PC but wasn't with v1.4.1. Happens with both Vulkan and OpenGL ES and with or without HD textures.

### Steps to Reproduce

1. Stare a new game
2.Do not skip the opening cut scene
3.  Proceed through dialogue between Link and Rusl
4. After the stand up, there is a wide shot with them and Epona where the music starts. The screen goes black and then transitions to them walking toward the bridge. The music cuts out as soon as the screen goes black.

### Expected Behavior

The music should continue along through the cut scene.

### Current Behavior

The music cuts out when the screen goes black before transitioning to the next part of the scene.

### OS / Architecture

Windows x86-64

### Game Disc

GameCube NTSC-U (North America)

### Game Language

English

### Graphics Backend

Vulkan

### Texture Packs and Mods

TP Definitive Edition+ 2.1

NotNite:
I also experienced this same issue within the ending credits; the music that plays during the cutscene in the middle of the credits works, though.

TheGag96:
Observed the particular issue mentioned in the issue marked as duplicate (#2534).

Issue #2516, opened 2026-09-19
King Bulblin victory music wont play


DendiLadd:
### Mods and Forks

- [x] The bug occurs on an **official release** of Dusklight, without any forks or third-party patches installed

### Dusklight Version

2.0.0
2.0.1

### Bug Description

When Link does his victory pose with Epona, there's absolutely no music. You can still hear sound effects like wind.

### Steps to Reproduce

1. Beat King Bulblin on bridge
2. Cutscene will play without music

### Expected Behavior

Music should play

### Current Behavior

Music not playing

### OS / Architecture

Windows x86-64

### Game Disc

GameCube NTSC-U (North America)

### Game Language

English

### Graphics Backend

Vulkan

### Texture Packs and Mods

None

### Additional Context

Looks like it might be related to issue #2503 and #2512 as both of their issues happened in my game when testing

tomlube:
this should be fixed in 2.0.1 hopefully?

DendiLadd:
Tested on 2,0.1 and the issue still exists :')

kamilink7:
It seems that this issue may affect multiple scenes, most notably the intro cutscene music, which cuts out after the first scene transition with and without Faster Scene Transitions on.

PR #2560, opened 2026-09-21
Corrected SOUND_TYPEID_STREAM_ALT to match previous 0x71 value

kamilink7:
Fixes issues #2503 and #2516 (audio streams queued before scene transitions do not play after scene change?). SOUND_TYPEID_STREAM (0x70) was set instead of (SOUND_TYPEID_STREAM_ALT) (0x71), which I noticed was the value before commit 67eea159 (JAudio doc/naming pass). Jaudio.md notes 0x71 as the value for audio streams that should fade/drop from queue after a scene transition, so this may have accidentally been causing the opposite effect, wherein tracks that are explicitly _not_ supposed to drop after a scene transition were being told by Z2SceneMgr to.


PJB3005:
Damnit

PJB3005 (approved):

PR #2580, opened 2026-09-22
Disable Easy Quick Spin in speedrun mode

999sian:
Easy Quick Spin is the only cheat `resetForSpeedrunMode()` doesn't neutralize, so holding R during an attack still forces a Quick Spin in a timed run.


999sian:
?

